PR #71, opened 2021-04-20
Updating react-native-qrcode-scanner

jf-morales:
`react-native-camera` was also updated.

Tested on IOS and Android (not in real devices)


geastwood:
@jf-morales can you resolve the conflict?

geastwood (approved):

PR #74, opened 2021-05-27
Fix IOS build

ahoseinian:
The app was not building using Xcode 12.5 because of some problems with flipper. 

the fix here uses these answers 
- https://github.com/facebook/react-native/issues/31179#issuecomment-829536845 (part 3)
- https://stackoverflow.com/a/67314652/1194644

### DIFF
--- a/ios/File.swift
+++ b/ios/File.swift
@@ -1,0 +1,8 @@
+//
+//  File.swift
+//  Litentry
+//
+//  Created on 26.05.21.
+//
+
+import Foundation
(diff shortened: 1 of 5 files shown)


geastwood:
Thanks for the PR.
There is currently a PR on upgrading RN to 0.64, would this PR still relevant after the upgrade?

geastwood:
I think it would be fine with 0.64, let's merge this.

geastwood (commented):
LGTM

ahoseinian:
@geastwood  nice, btw I created the develop branch here so maybe we can reserve the master to be merged when we want a new version of the app. sounds good?

geastwood:
Sounds good.

PR #78, opened 2021-05-31
Separate api loading from app screens

### LINKED ISSUE #77: Stuck on loading screen when api connection fails
currently, the user would get stuck on a loading page in case the selected API connection fails. also, there is no way for the user to change the selected server or go back to the previous state.

### proposed solution:
- add a dedicated screen for showing the connection state and  switch servers (`ApiLoadingNavigator`)
- main app screens would not be available unless we have a successful connection
- for switching server simply navigate to `ApiLoadingNavigator`

result:
https://user-images.githubusercontent.com/1168983/120181264-1b1a3d80-c20d-11eb-9b7b-95c687ec4b22.mp4

ahoseinian:
issue #77 

This PR solves the problem of getting stuck on the loading screen after a connection error by giving the user the possibility to switch to other networks

there are two major changes
- I removed `withNetworkSelect` and replaced it with a page that you can navigate to
- I created the new page under the Atomic/pages folder. I don’t know if you are a fan of that. I can change them to reflect the current folder structure of course.

the idea with the atomic structure is as follows:
atomic/
 - atoms -> small components btns, icons, texts & etc..
 - molecules -> component containing couple of atoms
 - organisms -> component containing molecules & atoms
 - pages -> navigation screens contains everything
 
also, I would normally don't break anything down into smaller pieces unless it'll be used in more than just one place in the app otherwise it can stay on the page file.

demo:
https://user-images.githubusercontent.com/1168983/120181264-1b1a3d80-c20d-11eb-9b7b-95c687ec4b22.mp4


src/Atomic/pages/ApiLoadingPage.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  /// navigate to app in case of successfully connecting
+  const prevApi = React.useRef(!!api);
+  React.useEffect(() => {
+    if (api && !prevApi.current) {
+      navigation.navigate('App');
+    }
+  }, [navigation, api]);
+
+  /////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
geastwood, line 28: what is this?

src/navigation/AppNavigator.tsx
```diff
@@ -34,9 +37,12 @@ function DashboardStackNavigator() {
   );
 }
 
+/////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
geastwood, line 40: let's remove this.

src/navigation/AppNavigator.tsx
```diff
@@ -55,4 +61,54 @@ function AppNavigator() {
...
```
geastwood, line 64: same here

src/navigation/AppNavigator.tsx
```diff
@@ -55,4 +61,54 @@ function AppNavigator() {
...
+        <AppStack.Screen name={'App'} component={DrawerNavigator} />
+      ) : undefined}
+      <AppStack.Screen name={'ApiNavigator'} component={ApiLoadingNavigator} />
+    </AppStack.Navigator>
+  );
+}
+
 export default AppNavigator;
+
+/////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
geastwood, line 84: and here

geastwood (approved):
Small things to remove and mergeable.

src/Atomic/pages/ApiLoadingPage.tsx
ahoseinian: I am used to separating stuff with a line of comment like this. I will remove them ✅

silva-fj (commented):
@ahoseinian It looks good to me 👍🏼 

> I created the new page under the Atomic/pages folder. I don’t know if you are a fan of that. I can change them to reflect the current folder structure of course.

Let's not introduce a new folder structure as part of this PR. We can open an issue and have a discussion (there are still some open topics about the project we should discuss). cc @geastwood

ahoseinian:
@franjs make sense I will refactor this to current format and then merge it

PR #83, opened 2021-06-01
Add ability to reconnect or switch server after disconnection

### LINKED ISSUE #62: Polkadot treasury motion #29 | Milestone-1 intake review | Stability, Connection to mainnet is getting last.
Inbetween using the App, connection to mainnet is getting lost, to recover have to reboot the App.
[Could be Stability]

**Validation Setup**

| Component | Version |
| --- | --- |
| Release APK | litentry-governance_v1.0.1.apk |
| Android Version | 7.1.1 |
| Litentry Devnet | wss://18.140.130.138 |
| Parity Signer | v4.5.3-beta |

ahoseinian:
fixes #62

- User would be redirected to apiLoadingScreen after losing connection
- User has the ability to reconnect or change network there

https://user-images.githubusercontent.com/1168983/120329850-8b4bc080-c2ec-11eb-8269-6c979e17783c.MP4

### DIFF
--- a/src/screen/ApiLoadingScreen.tsx
+++ b/src/screen/ApiLoadingScreen.tsx
@@ -5,7 +5,7 @@ import {
   TouchableOpacity,
   StyleSheet,
 } from 'react-native';
-import {Icon, Layout, Text} from '@ui-kitten/components';
+import {Button, Icon, Layout, Text} from '@ui-kitten/components';
 import globalStyles, {monofontFamily, colorGreen} from 'src/styles';
 import ScreenNavigation from 'layout/ScreenNavigation';
 import NetworkItem from 'presentational/NetworkItem';
@@ -18,8 +18,8 @@ export function ApiLoadingScreen({
 }: {
   navigation: NavigationProp<any>;
 }) {
-  const {currentNetwork} = useContext(NetworkContext);
-  const {api} = useContext(ChainApiContext);
+  const {currentNetwork, select} = useContext(NetworkContext);
+  const {api, inProgress} = useContext(ChainApiContext);
 
   /// navigate to app in case of successfully connecting
   const prevApi = React.useRef(!!api);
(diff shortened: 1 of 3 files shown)


silva-fj (approved):
Thank you for the demo! 👍🏼

src/screen/ApiLoadingScreen.tsx
```diff
@@ -45,31 +45,77 @@ export function ApiLoadingScreen({
...
-            style={[globalStyles.inlineIconDimension, {color: colorGreen}]}
-            name="planet"
-            pack="ionic"
-          />
-          <Text style={styles.text}>
-            Connecting to {currentNetwork?.name ?? ''}
-          </Text>
-        </View>
-        <ActivityIndicator size={'large'} color={colorGreen} />
+        {(() => {
```
geastwood, line 48: Don't understand why this is wrapped in a IIFE and the usage of `<>`

ahoseinian: well, I thought writing the conditions in if...else would be more readable than the ternary operator.
I could use a View with center alignment instead of <>, however, because the wrapping view was aligned already I used <> so the components instead would be direct children of that.

should I change this?

geastwood: Ah, I see. That mean, if you omit the `<>`, the alignment will break?

ahoseinian: you can't omit the `<>` [(React.Fragment)](https://reactjs.org/docs/fragments.html#short-syntax) because there are two children inside, that would be a JSX syntax error. I can replace it with a a `<View>` though but I have to style it so the alignment is right.

geastwood: True, didn't see that. haha

geastwood:
Go ahead, let's merge this.

PR #85, opened 2021-06-04
Feature/57 council screen

### LINKED ISSUE #57: Polkadot treasury motion #29 | Milestone-1 intake review | "Council" fragment navigation failure
Unable to navigate to "Council" fragment

After choosing "OK" no response

| Component | Version |
| --- | --- |
| Release APK | litentry-governance_v1.0.1.apk |
| Android Version | 7.1.1 |
| Litentry Devnet | wss://18.140.130.138 |
| Parity Signer | v4.5.3-beta |

ahoseinian:
Add the council page

https://user-images.githubusercontent.com/1168983/120766942-00044200-c51b-11eb-836c-18b8ae7eda2b.mp4

### DIFF
--- a/src/navigation/AppNavigator.tsx
+++ b/src/navigation/AppNavigator.tsx
@@ -63,7 +63,11 @@ const ApiLoadingStack = createStackNavigator();
 function ApiLoadingNavigator() {
   return (
     <ApiLoadingStack.Navigator headerMode={'none'} mode={'modal'} screenOptions={{gestureEnabled: false}}>
-      <ApiLoadingStack.Screen name={'ApiLoadingPage'} component={ApiLoadingScreen} options={{gestureEnabled: false}} />
+      <ApiLoadingStack.Screen
+        name={'ApiLoadingScreen'}
+        component={ApiLoadingScreen}
+        options={{gestureEnabled: false}}
+      />
       <ApiLoadingStack.Screen
         name={'NetworkSelectScreen'}
         component={NetworkSelectScreen}
--- a/src/navigation/AppNavigator.tsx
+++ b/src/navigation/AppNavigator.tsx
@@ -14,25 +14,18 @@ import * as routeKeys from 'src/navigation/routeKeys';
 import {ChainApiContext} from 'context/ChainApiContext';
 import {ApiLoadingScreen} from 'screen/ApiLoadingScreen';
 import {NetworkSelectScreen} from 'screen/NetworkSelectScreen';
+import {CouncilScreen} from 'screen/CouncilScreen';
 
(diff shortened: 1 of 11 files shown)


src/navigation/AppNavigator.tsx
```diff
@@ -79,15 +62,8 @@ export default AppNavigator;
...
-      headerMode={'none'}
-      mode={'modal'}
-      screenOptions={{gestureEnabled: false}}>
-      <ApiLoadingStack.Screen
-        name={'ApiLoadingPage'}
-        component={ApiLoadingScreen}
-        options={{gestureEnabled: false}}
-      />
+    <ApiLoadingStack.Navigator headerMode={'none'} mode={'modal'} screenOptions={{gestureEnabled: false}}>
+      <ApiLoadingStack.Screen name={'ApiLoadingPage'} component={ApiLoadingScreen} options={{gestureEnabled: false}} />
```
silva-fj: nit: We should probably drop `"Page" `to be consistent.. I got used to call them `pages` in Flutter as well :) but in RN `"screens"` are usually the standard.

src/screen/CouncilScreen.tsx
```diff
@@ -0,0 +1,97 @@
...
+        {loading ? (
+          <View style={globalStyles.centeredContainer}>
+            <Spinner />
+          </View>
+        ) : null}
+        <View style={styles.header}>
+          <Text category={'s1'}>Members</Text>
+          <Text category={'p2'}>{`seats ${data?.length}/${data?.length}`}</Text>
+        </View>
+        <List
```
silva-fj: Let's just use `FlatList` in this case, the `List` component is not adding anything and the type inference is pretty bad.

```
<FlatList
    data={data}
    renderItem={({item}) => {
        const text = u8aToString(item.info.display.asRaw);
         return <ListItem title={text} accessoryLeft={() => <Identicon value={item.accountId} size={30} />} />;
     }}
     keyExtractor={(item, index) => item.accountId.toString() ?? index.toString()}
     ItemSeparatorComponent={Divider}
 />

```

So we don't have to do: `{item: Exclude<typeof data, undefined>[0]}`

silva-fj (changes requested):
It looks good to me, just minor suggestions 👍🏼

src/navigation/AppNavigator.tsx
ahoseinian: done ✅

src/screen/CouncilScreen.tsx
ahoseinian: much better, thanks 👍

silva-fj (approved):

PR #128, opened 2021-06-15
Setting up global network. Fixes #64

silva-fj:
With these changes users will be able to add an account only to the network they are connected.

Other relevant changes:
- New AccountsContext: With this new context we have the infrastructure in place to add the multi account support.
- Fixes some unnecessary re-renders related to accounts and the current network.
- NetworkSelectScreen was removed in favor of a new NetworkSelect modal to be consistent with the network switch from the Dashboard and to have single way to switch networks.
- Some types are fixed.

**Demo**

https://user-images.githubusercontent.com/8374946/122091601-42762a80-ce09-11eb-84cc-8aff5a8341cf.mp4

PR #135, opened 2021-06-18
Feature/127 proposal votes

### LINKED ISSUE #127: Milestone 2 | Governance Actions | Motions
> As a user, I want to be able to vote or close a motion:

### acceptance criteria 
- [x] show all the available motions
- [x] motion params should be accessible by expanding the motion
- [x] user should be able to close a motion after the voting period is passed.
- [x] user should be able to vote during the voting period

ahoseinian:
### Description
Add motions screen to give users the ability to vote on an open motion or close a motion after the voting period is passed.

## Info
I had to use some of the functions from polkadot js library to get the necessary data and have left comment where we can find the source
- useVotingStatus
- useFormatBalance

Demo:

https://user-images.githubusercontent.com/1168983/122520925-84bf8780-d014-11eb-9741-74813121f4a3.MP4

fixes #127


geastwood:
Is everyone can vote on a `Motion`? I don't find relevant doc on this.

src/presentational/Account.tsx
```diff
@@ -0,0 +1,25 @@
+import React, {useContext} from 'react';
+import {ChainApiContext} from 'context/ChainApiContext';
+import useAsyncRetry from 'react-use/lib/useAsyncRetry';
+import {getAccountsIdentityInfo} from 'service/api/account';
+import {AccountId} from '@polkadot/types/interfaces';
+import {IdentityInfo} from '@polkadot/types/interfaces/identity/types';
+
+export function Account({
```
silva-fj: Should this component live in `src/layout/Account` ? to me presentational component it's a dumb component that takes props, no fetching data. 

Have you seen `src/layout/Address.tsx`? it looks like you are trying to accomplish a similar thing. I get the feeling that we are duplicating things that are already built in the project

src/presentational/Params.tsx
```diff
@@ -0,0 +1,74 @@
...
+import {standardPadding} from 'src/styles';
+import {useFormatBalance} from '../hook/useFormatBalance';
+
+export interface Param {
+  name: string;
+  type: TypeDef;
+  value?: Codec;
+}
+
+export function Params({data}: {data: Param[]}) {
```
silva-fj: The name and purpose of this component it's a bit confusing, `Params` is a generic name, but the component displays data related to an account, balance, etc. 

Is this meant to be reused?

src/screen/Council/CouncilScreen.tsx
```diff
@@ -69,5 +66,21 @@ export function CouncilScreen({navigation}: {navigation: NavigationProp<Dashboar
...
+    flexDirection: 'row',
+    justifyContent: 'flex-end',
+  },
+  listHeader: {justifyContent: 'space-between', flexDirection: 'row', padding: standardPadding * 3},
+  content: {paddingVertical: standardPadding, paddingHorizontal: standardPadding * 2},
 });
+
+export function useCouncilMembers() {
+  const {api} = useContext(ChainApiContext);
+  return useQuery('members', async () => {
```
silva-fj: Is there any other data we could have of members of other kind? How about `"council-members"` as queryKey?

src/screen/Council/MotionsScreen.tsx
```diff
@@ -0,0 +1,273 @@
...
+import Padder from 'presentational/Padder';
+import type {IExtrinsic, IMethod} from '@polkadot/types/types';
+import {Compact, GenericCall, getTypeDef} from '@polkadot/types';
+import {useCouncilMembers} from 'screen/Council/CouncilScreen';
+import {useQuery, useQueryClient} from 'react-query';
+import {Param, Params} from 'presentational/Params';
+
+export function MotionsScreen({navigation}: {navigation: NavigationProp<DashboardStackParamList>}) {
+  const {api} = useContext(ChainApiContext);
+  const {data, refetch, isLoading} = useQuery('motions', () => api?.derive.council.proposals());
```
silva-fj, line 21: Are proposals and motions the same? Why are we calling it Motions?

src/screen/Council/MotionsScreen.tsx
```diff
@@ -0,0 +1,273 @@
...
+        <Padder scale={0.5} />
+        {(() => {
+          if (isMember) {
+            if (isCloseable) {
+              return (
+                <View>
+                  <Button
+                    status={'warning'}
+                    size={'tiny'}
+                    onPress={() => {
```
silva-fj: nit: can we extract these functions to make the component easier to read?  e.g `onPress={onPressClose}`

src/screen/Council/CouncilScreen.tsx
```diff
@@ -69,5 +66,21 @@ export function CouncilScreen({navigation}: {navigation: NavigationProp<Dashboar
...
+  header: {
+    padding: standardPadding * 2,
+    flexDirection: 'row',
+    justifyContent: 'flex-end',
+  },
+  listHeader: {justifyContent: 'space-between', flexDirection: 'row', padding: standardPadding * 3},
+  content: {paddingVertical: standardPadding, paddingHorizontal: standardPadding * 2},
 });
+
+export function useCouncilMembers() {
```
silva-fj: We are getting the same data in the `CouncilScreen` without caching it, can you extract this to its own file in the hook folder and reuse it?

src/screen/Council/MotionsScreen.tsx
```diff
@@ -0,0 +1,273 @@
...
+  );
+}
+
+const styles = StyleSheet.create({flatList: {padding: standardPadding * 2}});
+
+function Motion({item}: {item: DeriveCollectiveProposal}) {
+  const theme = useTheme();
+  const {api} = useContext(ChainApiContext);
+  const {start} = useContext(TxContext);
+  const {accounts} = useContext(AccountContext);
```
silva-fj: after #128 gets merged, you need to change this to: 
```
...
import {useAccounts} from 'src/context/AccountsContext';
...

const {accounts} = useAccounts();
```

silva-fj (changes requested):
@geastwood @ahoseinian We should probably have a session to review the codebase together to check what we already have and what can be reused.. We have a lot of duplicated code

src/screen/Council/CouncilScreen.tsx
ahoseinian: 9c90f27 ✅

src/screen/Council/MotionsScreen.tsx
ahoseinian: 5703856 ✅

src/screen/Council/MotionsScreen.tsx
ahoseinian: it's a bit confusing, as far as I could understand it. a proposal is called a motion in https://polkadot.js.org/ when it's ready to be voted on. for example you can go to the treasury and submit a proposal there. but it wouldn't show up on this call until you specifically move it to the council. 

I think some proposals can become a referenda and go through public voting, which is different from this.

I picked `motions` as the query key because it's what this page is called here https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fkusama.api.onfinality.io%2Fpublic-ws#/council/motions which is showing this data

src/screen/Council/CouncilScreen.tsx
ahoseinian: yes, `council-members` would be the same as the hook name also, I'll change it.

src/presentational/Params.tsx
ahoseinian: yes, it's a bare implementation of this package. https://github.com/polkadot-js/apps/tree/master/packages/react-params/src/Param

the purpose is to read the params of a Call object and show info about it.

maybe we can have a `packages` folder and have this moved there for now before we complete it?

src/presentational/Account.tsx
ahoseinian: @franjs yes, you are right, seems the same. but I'd like to keep this version and replace the other, as this version just gives us the detail but we can change the style however we like in each place?
is that alright with you?

ahoseinian: I removed the `Address` in 4f495dd. we can revert that if not desired.

ahoseinian:
> Is everyone can vote on a `Motion`? I don't find relevant doc on this.

no you need to be a council member to vote on a motion.

src/presentational/Account.tsx
silva-fj: Yes that's fine, as long as we have what we need.. The idea is to reuse it unless there are many differences between them 👍🏼

src/presentational/Params.tsx
silva-fj: I see.. I guess we could have our own packages (if we cannot use them directly from the polkadot-js libraries), that's a good idea. Nevertheless, I think we should use names that reflect better what we are building (Features/UI).. As someone new to the code, If I want to build something where I need to display account and balance info for example, I would never guess that there is an UI element called `Params` that I can use for that xD. They use really bad names because they are based in data structures and no UI/Features

src/screen/Council/MotionsScreen.tsx
silva-fj: it really is a bit confusing, I need to read more about this :D

src/presentational/Account.tsx
silva-fj: Can you move it to `src/layout/Account` ?

ahoseinian: yes 7e02ec8 ✅

src/presentational/Params.tsx
ahoseinian: unfortunately, it seems they wrote their packages with only the web in mind. there is not a good separation between their views and logic which makes their code unusable for the most part for us. 😞

ahoseinian: 1b5e4cf moved to packages folder here. it's not a real package just inside a folder named packages and I called it `CallInspector` couldn't come up with a better name 😅 . maybe when we brushed up this and make sure it support all forms of params then extract it as a standalone package.

src/screen/TreasuryScreen.tsx
```diff
@@ -13,11 +13,13 @@ import Identicon from '@polkadot/reactnative-identicon';
...
 import {getAccountsIdentityInfo} from 'service/api/account';
 import {EmptyView} from 'presentational/EmptyView';
+import {useFormatBalance} from '../hook/useFormatBalance';
 
 export function TreasuryScreen({navigation}: {navigation: NavigationProp<DashboardStackParamList>}) {
+  const theme = useTheme();
   const {api} = useContext(ChainApiContext);
 
-  const {loading, value: treasuryData} = useAsyncRetry(async () => {
+  const {loading, value, retry} = useAsyncRetry(async () => {
```
silva-fj, line 22: Maybe we should create an issue to replace the usages of this hook in favor of useQuery from react-query for api related hooks

silva-fj (approved):

src/screen/TreasuryScreen.tsx
ahoseinian: #154 ✅

PR #111, opened 2021-06-09
Feature/59 submit tip

ahoseinian:
Simple page that runs `tip.reportAwesome` api. 

the transaction is not completing in my simulator maybe @geastwood can provide a way to test the signing process? [resolved]

https://user-images.githubusercontent.com/1168983/121381067-ae5a1e00-c945-11eb-9381-79dabe866087.mp4

> Note: this PR also tested with signing process.

PR #148, opened 2021-06-22
Feature/134 navigation header

ahoseinian:
### using react-navigation header sample.

this is a sample one-page change to react-navigation header instead of the ui-kitten version.
- I think the way react-navigation works is a bit more powerful and allow to change the header depending on which stack a route is on
- safe area view problem would be easier solved https://github.com/litentry/litentry-app/pull/111#discussion_r652590811
- component only handles the content 

on the downside
- component and header are separated so you have to look for it.


ahoseinian:
@franjs @geastwood I just changed one screen if you approve it will do the rest

silva-fj (dismissed):
It looks good 👍🏼

src/styles.ts
```diff
@@ -53,6 +53,7 @@ const globalStyles = StyleSheet.create({
     flex: 1,
   },
   icon: {...inlineIconDimension, ...iconColor},
+  icon30: {width: 25, height: 25},
```
silva-fj: What does `icon30` mean?

src/navigation/routeKeys.ts
```diff
@@ -1,14 +1,14 @@
 export const dashboard = 'Dashboard' as const;
-export const tips = 'Tips' as const;
-export const tipDetail = 'TipDetail' as const;
-export const submitTip = 'SubmitTipScreen' as const;
-export const registrarList = 'RegistrarList' as const;
-export const motionDetail = 'MotionDetail' as const;
-export const notificationSettingsScreen = 'NotificationSettingsScreen' as const;
+export const tipsScreen = 'Tips' as const;
```
silva-fj, line 2: NIT: Why only adding "screen" to this one? I think we should either drop or add "Screen" to all of them

src/navigation/routeKeys.ts
```diff
@@ -1,14 +1,14 @@
...
+export const tipDetail = 'Tip' as const;
+export const submitTip = 'SubmitTip' as const;
+export const registrarList = 'Registrars' as const;
+export const motionDetail = 'Motion' as const;
+export const notificationSettingsScreen = 'Notification' as const;
```
silva-fj, line 7:
```suggestion
export const notificationSettingsScreen = 'NotificationSettings' as const;
```

silva-fj (dismissed):

src/styles.ts
ahoseinian: changed to `icon25` it was a mistake.

src/navigation/routeKeys.ts
ahoseinian: I added `Screen` to all 👍

geastwood (approved):
Godspeed

PR #156, opened 2021-06-24
Feature/153 add push notification settings page

### LINKED ISSUE #153: Feature: Add Push notification setting page
> As **User** I'd like to
> **Choose different push topics**,
> to **stay informed** on the relevant events to me

IOS apps require a prompt

### Acceptance criteria:
- [ ] user able to change different topics they are interested in

### implementation details (if necessary)
Untitled

ahoseinian:
### Add Push Notification Setting page

current topics are [topics](https://www.notion.so/litentry/d661f3ac44d2421bbe15f080e28e7c08?v=a35178ed4a14404db7fa35518ed6c252)
```
const TOPICS = [
  {id: 'REF_VOTE', label: 'Vote on Active Referenda'},
  {id: 'COUNCIL_VOTE', label: 'Vote for Council Election'},
  {id: 'EMERGENCY_VOTE', label: 'Vote for Emergency Proposals'},
  {id: 'TREASURY_VOTE', label: 'Vote for Treasury Proposals'},
];
```


src/context/DataContext.tsx
```diff
@@ -9,6 +9,8 @@ export const DataContext = createContext<DataContextValueType>({
   },
 });
 
+export const useDataContext = () => useContext(DataContext);
```
silva-fj, line 12: I don't think we need this context, this can be a simple hook. Also, the library already have a hook that is doing the same (useAsyncStorage). I'll create an issue to refactor this

silva-fj: Issue: #158

src/navigation/navigation.d.ts
```diff
@@ -19,6 +19,7 @@ type DrawerParamList = {
   RegistrarList: undefined;
   Webview: {uri: string; title: string};
   DevScreen: undefined;
+  PushSettingsScreen: undefined;
```
silva-fj: How about `NotificationSettingsScreen` ?  It's kind of a nitpick

src/screen/DrawerScreen.tsx
```diff
@@ -178,6 +178,13 @@ function DrawerScreen({navigation}: DrawerContentComponentProps) {
             }
           />
           <Divider />
+          <ListItem
+            title="Push Settings"
```
silva-fj: Since this is actually something the user will see, I think `Notifications` or `Notification Settings` would be better. What do you guys think? @ahoseinian, @geastwood, @stephanie-olp, @lmdobreva

src/screen/PushSettingsScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+                )}
+              />
+            ))
+          )}
+          <Padder scale={2} />
+          <Divider />
+        </View>
+        <ListItem
+          accessoryLeft={(p) => <Icon {...p} name={'info-outline'} />}
+          title={"Don't forget to enable notifications in your phone setting in the app section"}
```
silva-fj:
```suggestion
          title={"Don't forget to enable notifications in your phone's settings"}
```

silva-fj: This only applies to IOS users, right?

src/screen/PushSettingsScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    padding: standardPadding * 3,
+  },
+});
+
+const TOPICS = [
```
silva-fj: I think the topics should come from the backend, so we don't have to maintain them in two places. Also:

- When we add support for new topics in the backend, the app gets them for free.
- If for some reason there are topics that are specific to a network, we don't have to handle that logic in the app as well.

For now it's fine, we can do this later

src/screen/PushSettingsScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+  {id: 'REF_VOTE', label: 'Vote on Active Referenda'},
+  {id: 'COUNCIL_VOTE', label: 'Vote for Council Election'},
+  {id: 'EMERGENCY_VOTE', label: 'Vote for Emergency Proposals'},
+  {id: 'TREASURY_VOTE', label: 'Vote for Treasury Proposals'},
```
silva-fj: We probably should use the chain event names as ids so we don't have to map them

silva-fj (dismissed):

src/screen/PushSettingsScreen.tsx
ahoseinian: in android, it's also possible to disable notifications on the app settings. so I guess both?

src/screen/PushSettingsScreen.tsx
ahoseinian: ref https://github.com/litentry/push-notification/issues/2

src/screen/PushSettingsScreen.tsx
ahoseinian: makes sense, can you add the event name to this issue https://github.com/litentry/litentry-app/issues/159 ?

src/navigation/navigation.d.ts
ahoseinian: done ✅

src/screen/PushSettingsScreen.tsx
silva-fj: Yes I know what you mean.. What I was referring to is that on IOS the user won't get PN until it manually approves the permissions, so if the user by mistake declines the permissions (the prompt we have) the message makes sense because it helps to know where to enable it again.. that's not the case on android, so with this message a user could think that for the PN to work he/she needs to go to the settings and enable it manually

ahoseinian: true, I think I will revise this on https://github.com/litentry/litentry-app/issues/160. 
the title change is pushed.

src/screen/DrawerScreen.tsx
stephanie-olp: I added "Push"  so the user won't mixed it up with email notifications but as it is written in the text bellow "Notifications" should be enough

ahoseinian: done ✅

silva-fj (approved):

PR #165, opened 2021-06-25
Link to treasury page after recieving a treasury deeplink

### LINKED ISSUE #161: Feature: Setup Handling push links
> As **User** I'd like to
> **be directed to a specific page when clicking on notification**,
> to **see relevant info**.

### ⚠️ what do we need?

- [ ] configuration https://reactnavigation.org/docs/configuring-links

### Acceptance criteria:

- [ ] test it works on Android
- [ ] test it works on IOS

ahoseinian:
fixes #161

### Description
- Adds linking ability to treasury screen.
- listens to push events or getInitialNotification to opens the desired link 

@franjs  for now, I am assuming our pushes will contain a data object like `{deeplink: 'litentry://treasury'}` we can change this if we decide on some other convention.


src/navigation/routeKeys.ts
```diff
@@ -12,3 +14,21 @@ export const myIdentity = 'MyIdentity' as const;
 export const councilScreen = 'CouncilScreen' as const;
 export const treasuryScreen = 'TreasuryScreen' as const;
 export const motionsScreen = 'MotionsScreen' as const;
+
+export const linking: LinkingOptions = {
+  prefixes: ['litentry://'],
+
+  config: {
```
geastwood, line 26: What is this here?

ahoseinian: this is how you define a path for your screen. [more info](https://reactnavigation.org/docs/configuring-links#mapping-path-to-route-names).

so for example, this config says if I use `linkTo('treasury')` or I receive a deeplink with the format `litentry://treasury` the screen `treasuryScreen` should be opened.

it also make sure that the navigation state is fine. so the route history would be `['App', 'Dashboard', 'Treasury']`

geastwood: What I worry about is that this become "untested" code, when we rename `App` to `SomethingElse`, would that be the case?

geastwood:
And the conflicts have to be solved.

src/navigation/AppNavigator.tsx
```diff
@@ -21,11 +21,15 @@ import {NotificationSettingsScreen} from 'screen/NotificationSettingsScreen';
...
 import globalStyles from 'src/styles';
-import {submitTipScreen} from 'src/navigation/routeKeys';
+import {useFirebase} from 'src/hook/useFirebase';
 
 const DashboardStack = createStackNavigator<DashboardStackParamList>();
 
 function DashboardStackNavigator() {
+  // needs to be called inside a navigator in order to
+  // have access to navigator context
+  useFirebase();
```
silva-fj, line 32: It would be safer to make useFirebase to throw an error if the context is undefined instead of the comments

src/navigation/routeKeys.ts
ahoseinian: yes make sense, I've made 'App' as a route key so the variable would be the reference and any change to that would change the linking also.

src/navigation/AppNavigator.tsx
ahoseinian: done ✅

geastwood (approved):
LGTM

PR #176, opened 2021-07-02
Feature/refactor chain api

ahoseinian:
### Description
Removes the unused code from chainApiContext in order to make it more simple.


src/layout/RegistrarList.tsx
```diff
@@ -45,14 +45,14 @@ function RegistrarList() {
...
-          renderItem={({item: registrar}) => {
+          renderItem={({item: registrar, index}) => {
             return (
               <View style={styles.registrarTeaserContainer}>
                 <RegistrarTeaser
                   key={registrar.account.toString()}
                   address={registrar.account.toString()}
                   fee={registrar.fee}
-                  index={registrar.index}
+                  index={index}
```
silva-fj: This change will make it depend on the order of the array and not the registrar index itself.. If someone changes the sorting the index will be incorrect

silva-fj (dismissed):
cool 👍🏼 

I just have the concern with the registrar index, but the rest looks good to me

src/layout/RegistrarList.tsx
ahoseinian: it was actually not a real index before also, it was the index of the map that was added to the element. https://github.com/litentry/litentry-app/pull/176/files#diff-94b39d441330c0bfb1c8d6c6713ec641959d0aa3425cfe624cce102fe66ef16bL12.
I will just add a todo to look into it later and merge this :)

silva-fj: It is the real registrar index, the thing is that the api does not return the index as a field but in the correct order

silva-fj: We should probably extract that into a function to make it descriptive or add some comments in the meantime

ahoseinian: I've changed it in 19d1b59. please check if you'd like that better.

silva-fj (approved):

PR #190, opened 2021-07-15
Add onNotificationOpenedApp to handle background notifications

ahoseinian:
### Description

- Add onNotificationOpenedApp to handle background notifications
- Compare networks on keys to prevent selecting a network twice because of different object refs


geastwood (approved):
LGTM

PR #195, opened 2021-07-16
Feature/189 turn on push notifications by default

### LINKED ISSUE #189: Turn on Push Notifications by default (iOS & Android)
> As a user, I don't want to bother about turning on the push notifications in the App PN setting page.

Currently all the toggle buttons are turned off by default for both iOS and Android users :

We would like to turn on all of them by default

ahoseinian:
### Description
- Toggle all notification on first app start to on for android
- Toggle all notifications on push permission allowed to on for ios
- Add hook to check if it's first app start
- Move notification hooks to `/hooks`


src/hook/useGrantPermission.ts
```diff
@@ -0,0 +1,58 @@
...
+  };
+
+  return {grant, setPermissionGrantingToShown, granted, isLoading};
+}
+
+function permissionAllowed(status: FirebaseMessagingTypes.AuthorizationStatus) {
+  return status === messaging.AuthorizationStatus.AUTHORIZED || status === messaging.AuthorizationStatus.PROVISIONAL;
+}
+
+export function useShouldShowPushPermissionScreen() {
```
silva-fj: We don't need all this logic, to show the screen we just need to check `await messaging().hasPermission() === messaging.AuthorizationStatus.NOT_DETERMINED`

src/hook/useGrantPermission.ts
```diff
@@ -0,0 +1,58 @@
...
+    return Platform.OS === 'ios' && permissionAllowed(await messaging().hasPermission());
+  });
+
+  const {mutate: grant} = useMutation(async () => {
+    const status = await messaging().requestPermission();
+    const allowed = permissionAllowed(status);
+
+    if (allowed) {
+      for (const topic of topics) {
+        await toggleTopic({id: topic.id, subscribe: true});
```
silva-fj: let's separate the logic of requesting permissions and subscribing to topics, this can be done in the `usePushTopics` hook.

src/hook/useGrantPermission.ts
```diff
@@ -0,0 +1,58 @@
...
+import {Platform} from 'react-native';
+import {usePushTopics} from './usePushTopics';
+
+const PERMISSION_GRANTING_SCREEN_SHOWN = 'PERMISSION_GRANTING_SCREEN_SHOWN';
+
+export function useGrantPermission(options?: UseMutationOptions<boolean, unknown, void, unknown>) {
+  const {toggleTopic, topics} = usePushTopics();
+  const queryClient = useQueryClient();
+
+  const {data: granted, isLoading} = useQuery('permission_granted', async () => {
```
silva-fj: useQuery is not adding any value in this case and we don't need to check for Platform.OS, since on android permissions will be always granted

src/hook/useGrantPermission.ts
```diff
@@ -0,0 +1,58 @@
+import {useQueryClient, useQuery, useMutation, UseMutationOptions} from 'react-query';
+import * as AsyncStorage from 'src/service/AsyncStorage';
+import messaging, {FirebaseMessagingTypes} from '@react-native-firebase/messaging';
+import {Platform} from 'react-native';
+import {usePushTopics} from './usePushTopics';
+
+const PERMISSION_GRANTING_SCREEN_SHOWN = 'PERMISSION_GRANTING_SCREEN_SHOWN';
+
+export function useGrantPermission(options?: UseMutationOptions<boolean, unknown, void, unknown>) {
```
silva-fj: We can simplify the API a lot by having this hook to return something like: 

`const {hasPermissions, requestPermissions} = usePushNotificationsPermissions();`

Where `hasPermissions` is just a boolean that is `true` when the permissions is granted

src/hook/useGrantPermission.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (allowed) {
+      for (const topic of topics) {
+        await toggleTopic({id: topic.id, subscribe: true});
+      }
+    }
+
+    return allowed;
+  }, options);
+
+  const setPermissionGrantingToShown = async () => {
```
silva-fj: We don't need this

src/hook/useTurnOnAllNotificationsOnAppStartForAndroid.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function useTurnOnAllNotificationsOnAppStartForAndroid() {
+  const {toggleTopic, topics, isLoading: topicsLoading} = usePushTopics();
+  const {data: isFirstAppStart, isLoading: firstAppStartLoading} = useIsFirstAppStart();
+
+  const isRunnedOnceRef = React.useRef(false);
+
+  const isLoaded = !firstAppStartLoading && !topicsLoading && isFirstAppStart;
+  const shouldToggle =
+    isLoaded && isFirstAppStart && Platform.OS === 'android' && !isRunnedOnceRef.current && topics.length > 0;
```
silva-fj: `isLoaded` is already checking `isFirstAppStart`

src/hook/useTurnOnAllNotificationsOnAppStartForAndroid.ts
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import {Platform} from 'react-native';
+import {usePushTopics} from './usePushTopics';
+import {useIsFirstAppStart} from 'src/hook/useIsFirstAppStart';
+
+export function useTurnOnAllNotificationsOnAppStartForAndroid() {
+  const {toggleTopic, topics, isLoading: topicsLoading} = usePushTopics();
+  const {data: isFirstAppStart, isLoading: firstAppStartLoading} = useIsFirstAppStart();
+
+  const isRunnedOnceRef = React.useRef(false);
```
silva-fj, line 10: Why is this ref needed? isn't `isFirstAppStart` enough?

src/hook/useGrantPermission.ts
ahoseinian: We also need a way to check if user clicked on skip here. otherwise, it would pop up on every app start. 
Screenshot 2021-07-19 at 13 05 31

how would you think is the best way to do that?

src/hook/useGrantPermission.ts
ahoseinian: `granted` & `grant` are exactly that. but I guess the naming is a bit confusing I will change it to the version you requested.

src/hook/useGrantPermission.ts
ahoseinian: `messaging().hasPermission()` is  an async call. so we need to either use `useEffect` + `useState` to check that or we can use the `useQuery` hook. personally I like this better, do you think the other way is better?

src/hook/useGrantPermission.ts
silva-fj: I would remove the Skip button.. Apps normally just prompt the user to allow or deny permissions, there is no option to skip

ahoseinian: yes that will fix everything. I will do that.

src/hook/useGrantPermission.ts
silva-fj: I know `useQuery` is convenient here but it's a bit of an overkill and also it has a different purpose. I think we can add hook to handle simple cases like this one.

The API could be something like: 
`useAsync<T, E = Error>(fn: () => T | Promise<T>) => { isLoading: boolean; data: T; error: E }`

ahoseinian: yes works for me 👍

ahoseinian: I removed and refactored this can you take another look here?

src/hook/useGrantPermission.ts
ahoseinian: done ✅

src/hook/useTurnOnAllNotificationsOnAppStartForAndroid.ts
ahoseinian: `isFirstAppStart` stays true the whole time user has opened the app for the first time. I did it like that, so it could be used in different hooks if needed. therefor, if I don't do this check toggling the topics would cause the effect to run again and do the same thing over and over.

silva-fj (dismissed):

src/hook/useGrantPermission.ts
silva-fj: The new hook should be used here :)

src/hook/usePushNotificationsPermissions.ts
```diff
@@ -0,0 +1,46 @@
+import {useQueryClient, useQuery, useMutation} from 'react-query';
+import messaging, {FirebaseMessagingTypes} from '@react-native-firebase/messaging';
+import {usePushTopics} from './usePushTopics';
+
+export function usePushNotificationsPermissions() {
+  const {subscribeToAllTopics} = usePushTopics();
+  const queryClient = useQueryClient();
+
+  const {pushAuthorizationStatus} = usePushAuthorizationStatus();
```
ahoseinian, line 9: @franjs I changed the previous logic to use the  `usePushAuthorizationStatus`. it's not using  useAsync there but that's because I like the invalidate it after the mutation so user would be automatically moved to next route.

src/hook/useGrantPermission.ts
ahoseinian: https://github.com/litentry/litentry-app/pull/195#discussion_r672409965

silva-fj (approved):

PR #196, opened 2021-07-16
Starting to organize Api data related into it's own module

### LINKED ISSUE #162: Organize api data related calls into hooks
We have multiple ways to get data from the api (chain) in different places, and duplicated calls to get the same data. Having a single source of truth to get this data makes it easier to maintain and will help developers to identify what's currently available as well as guide them on how to get data from the chain.

silva-fj:
These are the initial steps to api related calls, the goal is having a single source of truth and a single way to interact with the api across the code base that is consistent and easier to follow for new contributors.

Relevant changes: 
- New api module was added (could be easily reused with the web app at some point if we want/need to.)
- useApiQuery hook added to abstract the logic of having the api always defined and usage of useQuery from react-query.

These changes don't include querying data from the api using subscriptions (this should be tackled in another PR) neither transactions Tx (as mutations). 

Fixes #162


ahoseinian (approved):
LGTM

PR #219, opened 2021-07-26
Show validation error on bad beneficiary value

### LINKED ISSUE #213: Validate beneficiary on "Submit Tip" form
When submitting a Tip, if the account of the beneficiary is not valid it silently fails giving no feedback to the user.

e.g: 
Screenshot 2021-07-23 at 2 34 30 PM
 
The following error is silenced: 

`Error: createType(Call):: Call: failed decoding tips.reportAwesome:: Struct: failed on args: {"reason":"Bytes","who":"AccountId"}:: Struct: failed on who: AccountId:: Decoding Klkkljjkljkljkl: Invalid base58 character "l" (0x6c) at index 1 `

ahoseinian:
### Description

Check if api return a validation error and show it to the user:

fixes #213

### DIFF
--- a/src/screen/SubmitTipScreen.tsx
+++ b/src/screen/SubmitTipScreen.tsx
@@ -56,7 +56,6 @@ export function SubmitTipScreen({navigation}: {navigation: NavigationProp<Dashbo
           disabled={!valid}
           onPress={() => {
             if (api && state.account) {
-              dispatch({type: 'SET_ERROR', payload: undefined});
               start({
                 api,
                 address: state.account,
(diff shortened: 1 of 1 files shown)


src/screen/SubmitTipScreen.tsx
```diff
@@ -55,6 +56,7 @@ export function SubmitTipScreen({navigation}: {navigation: NavigationProp<Dashbo
           disabled={!valid}
           onPress={() => {
             if (api && state.account) {
+              dispatch({type: 'SET_ERROR', payload: undefined});
```
silva-fj: It would be better to dispatch after `start` resolves (`then` block). 
How about resetting the entire state?

ahoseinian: actually, I think it's unnecessary. so I removed it 7c855a3. in case the promise succeed we do navigation.goBack which will clear the state. and in case it throws we dispatch in catch block. 
also I am removing the error as soon as user changes the input in beneficiary field.

silva-fj (approved):

PR #231, opened 2021-07-28
Feature/216 add missing info on referenda

### LINKED ISSUE #216: Missing info in the Referenda
We need to check the logic we are using to display the data for the referenda, currently the app does not display the same info as the polkadotjs app.

**Litentry App:**
Screenshot 2021-07-23 at 3 17 57 PM

_Code reference:_
https://github.com/litentry/litentry-app/blob/03b9e9234f55516f5150381aa77d3e8ca468a6ae/src/screen/ReferendaScreen.tsx#L50

**Polkadotjs App:**
Screenshot 2021-07-23 at 3 09 55 PM

### LINKED ISSUE #229: Bug: Spelling mistake on referendum page
It's written "Has of the proposal" instead of "Hash of the proposal"

### LINKED ISSUE #230: Bug: Wrong message when referendum page loading
As a user I expect to get the right message when something is loading.

Right now, when you tap the referenda section on the dashboard, the user has to wait more than 3 seconds for the right information to be shown. Meanwhile the following error message is displayed: 

That could lead the user to misunderstand the situation and quit the screen. If we can't improve the loading speed of this screen it would be nice to display a loading animation like this 
image

ahoseinian:
### Description
- Show preimage when proposal is missing
- Wrap theme provider with accounts provider so accounts would be available in modal

fixes #229 & #216
Screenshot 2021-07-28 at 11 45 22
Screenshot 2021-07-28 at 11 45 48


src/screen/ReferendaScreen.tsx
```diff
@@ -21,37 +22,40 @@ export function ReferendaScreen() {
...
-          }}
-          ItemSeparatorComponent={Divider}
-          keyExtractor={(item) => item.index.toString()}
-          ListEmptyComponent={EmptyView}
-        />
+        {isLoading ? (
+          <LoadingView />
+        ) : (
+          <FlatList
+            refreshing={isLoading}
```
silva-fj: This will never be true

silva-fj (dismissed):

src/screen/ReferendaScreen.tsx
ahoseinian: changed with `isFetching`

silva-fj (dismissed):

silva-fj (approved):

PR #233, opened 2021-07-28
Show dashboard with/without user

ahoseinian:
### Description
Show dashboard with/without user


silva-fj (approved):

geastwood:
So now user can only access "add account" at sidebar?

ahoseinian:
yes, we add that somewhere in new dashboard design if you'd like

ahoseinian:
@stephanie-olp ⬆️

PR #236, opened 2021-07-30
Go to next step after successful scan

### LINKED ISSUE #227: Bug: When scanning a QR Code, Picture should freeze
As a user, when I successfully scan QR Code, I expect to have a clear signal

For both iOS and Android Users
## Steps To Reproduce

1. Tap on Add an Account
2. Select "Via QR"
3.  Scan the QR Code of one your Polkadot/Kusama account

## The current behavior
When the phone detects a Qr Code it vibrates, but the camera keep rolling 

## The expected behavior
When the phone detects a Qr Code it can vibrate but more importantly the camera should freeze , otherwise it can be confusing for the user

ahoseinian:
fixes #227

Go to next step after successful scan of QR code


silva-fj (approved):

geastwood:
Rebase and mergeable. 👯

PR #237, opened 2021-07-30
Remove useAccountDetail in favor of useAccountIdentityInfo

ahoseinian:
- Remove useAccountDetail and replace it with useAccountIdentityInfo


src/api/hooks/useAccountIdentityInfo.ts
```diff
@@ -2,12 +2,15 @@ import {ApiPromise} from '@polkadot/api';
 import useApiQuery from 'src/api/hooks/useApiQuery';
 import {getAccountIdentityInfo} from 'src/api/queryFunctions/getAccountIdentityInfo';
 
-export function useAccountIdentityInfo(accountId: string) {
+export function useAccountIdentityInfo(accountId?: string) {
   return useApiQuery(
     ['account_identity', accountId],
     (api: ApiPromise) => {
+      if (!accountId) {
+        throw new Error('accountId is required');
```
silva-fj: if the query is not `enabled` this will never run

src/layout/AccountTeaser.tsx
```diff
@@ -29,10 +28,9 @@ function AccountTeaser(props: PropTypes) {
   const [qrVisible, setQrVisible] = useState(false);
   const {theme} = useTheme();
   const {accounts} = useAccounts();
-  const {currentNetwork} = useContext(NetworkContext);
-  const {api} = useContext(ChainApiContext);
   const account = accounts[0]; // TODO: change this when adding multi account support
```
silva-fj, line 30: are we still using this?

src/screen/AddAccountScreen/AddressPreview.tsx
```diff
@@ -28,8 +28,8 @@ function AddressInfoPreview(props: PropTypes) {
   const [account, setAccount] = useState<AccountInfo>();
   const [inProgress, setInProgress] = useState(false);
 
-  const {display, detail} = useAccountDetail(network?.key, address, api);
-  const identity = detail?.data;
+  const {data, isLoading} = useAccountIdentityInfo(address);
```
silva-fj: `isLoading` is not used

src/screen/AddAccountScreen/AddressPreview.tsx
```diff
@@ -28,8 +28,8 @@ function AddressInfoPreview(props: PropTypes) {
...
+  const identity = data?.registration;
 
   useEffect(() => {
     if (api) {
```
silva-fj: This effect, the `account` state and the `api` are not needed anymore, we are getting the account info from the new hook

src/screen/MyIdentityScreen/index.tsx
```diff
@@ -1,42 +1,38 @@
...
 
   const content = useMemo(() => {
-    if (inProgress || !address) {
+    if (isLoading || !address) {
       return <LoadingView />;
     }
 
-    if (detail?.data?.judgements.length) {
+    if (data?.registration?.judgements.length) {
       // there is already judgements to display
```
silva-fj: We could get info like this from `useAccountIdentityInfo` so it's explicit and we don't have to add comments.
e.g: `hasJudgements`, `hasIdentity`

src/api/hooks/useAccountIdentityInfo.ts
ahoseinian: yes I know, I just did it to fix the TS error since `getAccountIdentityInfo` expects an accountId. is there a better way to do this?

src/layout/AccountTeaser.tsx
ahoseinian: The file is actually not used anywhere, I just kept it because thought maybe it'll be used in future. should I remove it entirly?

src/api/hooks/useAccountIdentityInfo.ts
silva-fj:
```
if (accountId) {
    return getAccountIdentityInfo(api, accountId);
}
```

src/layout/AccountTeaser.tsx
silva-fj: I see.. Let's create an issue to remove it later if won't use it (I think at some point we need to revisit the accounts list)

src/screen/AddAccountScreen/AddressPreview.tsx
ahoseinian: I updated the PR and added a new hook called `useAccountInfo` here. problem is `identityInfo` ( Registration ) is different from `AccountInfo`. please take another look here.

src/api/hooks/useAccountIdentityInfo.ts
ahoseinian: makes sense :D

src/screen/MyIdentityScreen/index.tsx
ahoseinian: you mean something like 694b50a ?

src/api/queryFunctions/getAccountIdentityInfo.ts
```diff
@@ -17,13 +39,20 @@ export async function getAccountIdentityInfo(api: ApiPromise, accountId: string)
     const superRegistration = superAccountOption.unwrapOr(undefined);
 
     if (superRegistration) {
+      console.log('AAAA', u8aToString(superRegistration.info.display.asRaw));
```
silva-fj: You forgot to remove it

silva-fj (dismissed):
👍🏼

src/api/queryFunctions/getAccountIdentityInfo.ts
ahoseinian:
```suggestion
```

silva-fj (approved):

PR #246, opened 2021-08-03
Feature/improve api loading

### LINKED ISSUE #234: Research on solutions for "Keep getting disconnected from Network"
The user keeps getting disconnected from the Polkadot/Kusama network .I think Francisco already mentioned this issue but it's particularly annoying when user taps on a push notification and lands on this screen:

  
Once he clicks Reconnect, he does land on the right screen.

Do we have any way on our side to fix this problem ?

ahoseinian:
fixes #234 

- Implement automatic reconnection to api
- Remove api loading screen
- Updates push notification handling and changing of the network


src/screen/DeeplinkNavigatorScreen.tsx
```diff
@@ -0,0 +1,34 @@
+import {NavigationProp, RouteProp, useLinkTo} from '@react-navigation/native';
+import {ChainApiContext} from 'context/ChainApiContext';
+import {NetworkContext} from 'context/NetworkContext';
+import LoadingView from 'presentational/LoadingView';
+import React, {useContext} from 'react';
+import {AppStackParamList} from 'src/navigation/navigation';
+import {deeplinkNavigatorScreen} from 'src/navigation/routeKeys';
+
+// TODO: Remove this file and move the logic to push notification handling function
```
ahoseinian, line 9: I couldn't test this properly without a build. after the build I can fix this point in another PR.

src/api/hooks/useApiQuery.ts
```diff
@@ -21,7 +21,10 @@ function useApiQuery<TData>(
       }
       return queryFn(api);
     },
-    options,
+    {
+      ...options,
+      enabled: options?.enabled ? options.enabled && api?.isConnected : api?.isConnected,
```
silva-fj: This will work fine if we pass `enabled: true` or if we don't pass `enabled` at all, but if we pass `enabled: false` and the api is connected it will execute the query function when it shouldn’t. 

One way to solve this could be: 

1. Add default value: 
      `options: UseQueryOptions<TData> = { enabled: true },`
2. then 
    `enabled: options.enabled && Boolean(api?.isConnected)`

ahoseinian: done ✅

src/context/ChainApiContext.tsx
```diff
@@ -1,73 +1,60 @@
...
+    const provider = new WsProvider(wsAddress, false);
     const apiPromise = new ApiPromise({provider});
+    apiPromise.connect();
+    dispatch({type: 'CONNECT', payload: apiPromise});
 
     function handleConnect() {
       logger.debug('ChainApiContext: Api connected');
-      setStatus('connected');
+      dispatch({type: 'ON_CONNECT'});
+      queryClient.clear();
```
silva-fj, line 42: We could also add the network as prefix to the querykeys so we don't lose the cached data when changing the network. But it's out of scope for this issue, we can do it later

src/screen/DashboardScreen.tsx
```diff
@@ -77,12 +77,14 @@ export function DashboardHeaderLeft() {
...
   const {currentNetwork} = useContext(NetworkContext);
-  const {status} = useContext(ChainApiContext);
+  const {status} = useApi();
 
   return (
     <TouchableOpacity style={styles.titleContainer} onPress={() => setNetworkSelectOpen(true)}>
       <Text category="s1">Litentry</Text>
-      {currentNetwork ? <NetworkItem item={currentNetwork} isConnected={status === 'ready'} /> : null}
+      {currentNetwork ? (
+        <NetworkItem item={currentNetwork} isConnected={status === 'connected' || status === 'ready'} />
```
silva-fj, line 86: We could do this check in the hook if you think we would need to check this in other places:

`const {isReady} = useApi();`

src/screen/DeeplinkNavigatorScreen.tsx
silva-fj: Let's create an issue so we don't forget

src/screen/DeeplinkNavigatorScreen.tsx
```diff
@@ -0,0 +1,34 @@
...
+export function DeeplinkNavigatorScreen({
+  route,
+}: {
+  navigation: NavigationProp<AppStackParamList, typeof deeplinkNavigatorScreen>;
+  route: RouteProp<AppStackParamList, typeof deeplinkNavigatorScreen>;
+}) {
+  const linkTo = useLinkTo();
+  const {currentNetwork, select, availableNetworks} = useContext(NetworkContext);
+  const {api} = useContext(ChainApiContext);
```
silva-fj: what about `useApi`?

src/context/ChainApiContext.tsx
ahoseinian: Sounds good, we also have to make sure not to show very old data. (a refetch should happen while we show the cached data)

src/screen/DashboardScreen.tsx
ahoseinian: For now it's the only use case, I will keep this in mind if in future we were repeating this somewhere else.

src/screen/DeeplinkNavigatorScreen.tsx
ahoseinian: done ✅

src/screen/DeeplinkNavigatorScreen.tsx
ahoseinian: #258

silva-fj (approved):

PR #263, opened 2021-08-09
Feature/252 submit a proposal

### LINKED ISSUE #252: Feature: Governance Actions | Submit a proposal
> As a dots holder I want to submit a proposal so it will be added to the queue for future referendums.

The user should be able to submit a proposal from the democracy screen : 

When he clicks on "Submit a proposal", the following dialog box opens: 

 The user  should be able to do the following actions:
* Pick the account he wanna vote with
* Enter the Preimage hash
* Pick the locked balance

And he should see the following information:
* The minimum deposit

Once all the field were completed, he can either:
* Submit the proposal
Or
* Cancel and close the dialog box

If the submission is a success, the user should see a successful message saying "Congratulations, your proposal has been submitted ! "

If there is an error the following message is shown: "Your proposal could not be submitted. Please review your details."

ahoseinian:
Fixes #252 

- Add submit proposal button
- Figure out the TxContext details from transaction meta data


src/context/TxContext/index.tsx
```diff
@@ -64,16 +63,20 @@ function TxContextProvider({children}: PropTypes): React.ReactElement {
 
     modalRef.current?.open();
 
+    const {meta} = transaction.registry.findMetaCall(transaction.callIndex);
+    const args = meta?.args.map(({name}) => name).join(', ') || '';
+    const title = `Sending transaction ${section}.${method}(${args})`;
+    const description = formatCallMeta(meta);
```
silva-fj, line 69: What does the output of this look like? is it user friendly? Is it displayed somewhere?

ahoseinian: yes this is the same as before for transactions, but instead of giving it explicitly it is derived from the api. it looks like this currently in the app.
Screenshot 2021-08-10 at 10 26 40

silva-fj (dismissed):

silva-fj (approved):

PR #308, opened 2021-08-20
Adding useApiTx hook for starting transactions

### LINKED ISSUE #307: Refactoring transaction logic
Encapsulate the logic needed to make transactions into a hook.

silva-fj:
Adding new hook to be able to start transactions without having to always get the api and start from the respective contexts.

I thought we could clean a little more code but in several places we still need the api. However, the logic is still improved a little, later we can revisit the logic when we integrate with other methods to sign the transactions.

Fixes #307

### DIFF
--- a/src/api/hooks/useApiTx.ts
+++ b/src/api/hooks/useApiTx.ts
@@ -1,6 +1,7 @@
 import {useContext} from 'react';
 import {useApi} from 'context/ChainApiContext';
 import {TxContext, StartConfig} from 'context/TxContext';
+import {ApiPromise} from '@polkadot/api';
 
 type TxConfig = Omit<StartConfig, 'api'>;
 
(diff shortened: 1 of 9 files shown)


src/api/hooks/useApiTx.ts
```diff
@@ -0,0 +1,17 @@
...
+import {useApi} from 'context/ChainApiContext';
+import {TxContext, StartConfig} from 'context/TxContext';
+
+type TxConfig = Omit<StartConfig, 'api'>;
+
+export function useApiTx() {
+  const {start} = useContext(TxContext);
+  const {api} = useApi();
+
+  return (config: TxConfig) => {
```
ahoseinian: if we want to pass the api to use as params we could also do something like this I guess?
```suggestion
  return (config: (api: ApiPromise) => TxConfig) => {
```
and then use like
```
start((api) => ({params: api.....}))
```

ahoseinian (dismissed):

src/api/hooks/useApiTx.ts
silva-fj: Sounds good, I'll make it optional so we don't have to pass a function for simple cases

ahoseinian (approved):

PR #315, opened 2021-08-23
Feature/297 motion detail page

### LINKED ISSUE #297: Bug: Motion detail page loading after a minute
it takes very long time to load this page because we are using an external api from polkascan.io which is defect. and also it wouldn’t work on litentry network because polkascan doesn’t support that. maybe we can get all data required from chain.

- [x] See if we can remove the dependency to the polkascan.io for loading this page.

ahoseinian:
Fixes #297 
- Remove polkascan api and use the info on chain
- Don't show polkascan webview for litentry chain ( polkascan doesn't support that )
- Some jsx refactoring


src/presentational/AccountInfoInlineTeaser.tsx
```diff
@@ -27,6 +27,7 @@ function AccountInfoInlineTeaser({display, judgements}: PropTypes) {
 const styles = StyleSheet.create({
   text: {fontFamily: monofontFamily, fontWeight: 'bold'},
   container: {
+    flex: 1,
```
silva-fj, line 30: Can you please check the screens where we use this component that is not breaking with this change? I remember adding `flex: 1` here was breaking the UI somewhere. (Also check the usage of the AddressInlineTeaser)

src/screen/MotionDetailScreen.tsx
```diff
@@ -51,19 +51,35 @@ export function MotionDetailScreen(props: PropTypes) {
 
   const {data: motion} = useMotionDetail(params);
 
+  const {data} = useCouncilMembers();
+  const membersCount = data?.members.length ?? 0;
+  const {isCloseable, isVoteable, hasFailed, hasPassed} = useVotingStatus(motion?.votes, membersCount, 'council');
```
silva-fj: How about getting the status from this hook and remove all the logic down below?

```suggestion
  const {status} = useVotingStatus(motion?.votes, membersCount, 'council');
```

src/screen/MotionDetailScreen.tsx
```diff
@@ -51,19 +51,35 @@ export function MotionDetailScreen(props: PropTypes) {
...
+    ? 'Closed'
+    : hasPassed
+    ? 'Passed'
+    : 'Open';
+
   if (!motion) {
     return <LoadingView />;
   }
 
+  const proposer = motion.votes?.ayes[0];
```
silva-fj, line 62: Is the proposer always the first element?

src/presentational/AccountInfoInlineTeaser.tsx
ahoseinian: I checked everything seems to be fine now, not sure were was the problem before.

src/screen/MotionDetailScreen.tsx
ahoseinian: TBH i am not sure! in polkadot it doesn't show the proposer on motion page so I didn't have example code to double check. however, from what I checked. when somebody makes a new motion he/she will be automatically the first aye vote. that's where this came from

src/screen/MotionDetailScreen.tsx
ahoseinian: 8e333d1

src/screen/MotionDetailScreen.tsx
silva-fj: I wonder if they don't show it for a reason.. Let's create an issue to revisit this so we don't forget

silva-fj (approved):

src/screen/MotionDetailScreen.tsx
ahoseinian: yes sounds good. https://github.com/litentry/litentry-app/issues/321

Issue #334, opened 2021-08-30
Bug: Account Teaser not displaying sub accounts correctly


silva-fj:
Sub accounts should display the parent account name as well as the sub-identity name as shown here: 

<img width="447" alt="Screenshot 2021-08-30 at 9 42 58 AM" src="https://user-images.githubusercontent.com/8374946/131304408-a83ac720-a53c-4eed-8442-1b50c3c8bef1.png">

Currently the teaser is only displaying the parent account name:

<img width="417" alt="Screenshot 2021-08-30 at 9 41 03 AM" src="https://user-images.githubusercontent.com/8374946/131304500-40a79b82-0d4b-4096-aec9-95ac03ee65b1.png">

This is a regression bug, it was introduced in this [PR](https://github.com/litentry/litentry-app/pull/237)

This is what the old hook (`useAccountDetail`) was returning:

```
const displayFull = subAccountDisplay ? `${display}/${subAccountDisplay}` : display;
```

geastwood:
@ahoseinian can you please take a look, and revert some of the code in the old hook.

ahoseinian:
@geastwood will do 👍

PR #335, opened 2021-08-30
Set Sub accounts identity

### LINKED ISSUE #320: Add Sub Accounts Identity
Users can also link accounts by setting "sub accounts", each with its own identity, under a primary account. The system reserves a bond for each sub account.

More info: https://wiki.polkadot.network/docs/learn-identity#sub-accounts

silva-fj:
Adding the option to set sub identities. Fixes #320 

Screenshot 2021-08-30 at 4 44 20 PM
Screenshot 2021-08-30 at 4 02 57 PM
Screenshot 2021-08-30 at 4 03 18 PM
Screenshot 2021-08-30 at 4 40 03 PM
Screenshot 2021-08-30 at 4 23 58 PM

The sub accounts are not displayed correctly in the multi account view. There is already an issue for this bug #334


geastwood (dismissed):
LGTM

geastwood:
Mergeable only after rebase.

src/screen/subIdentities/RegisterSubIdentitiesScreen.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const {data: identityInfo} = useAccountIdentityInfo(address);
+  const {data: subIdentitiesData} = useSubIdentities(address);
+  const [subIdentities, setSubIdentities] = useState(subIdentitiesData || []);
+  const [submitSubsDisabled, setSubmitSubsDisabled] = useState(true);
+  const queryClient = useQueryClient();
+  const startTx = useApiTx();
+
+  useEffect(() => {
+    if (subIdentitiesData?.length) {
+      setSubIdentities(subIdentitiesData);
```
ahoseinian, line 36: I don't understand why we are not just using `subIdentitiesData` instead of having `subIdentities`?

src/screen/subIdentities/RegisterSubIdentitiesScreen.tsx
```diff
@@ -0,0 +1,182 @@
...
+  useEffect(() => {
+    navigation.setOptions({
+      headerRight: () => (
+        <TopNavigationAction
+          icon={(props) => <Icon {...props} name="plus-circle-outline" />}
+          onPress={onAddSubIdentityPress}
+        />
+      ),
+    });
+  });
```
ahoseinian:
```suggestion
  }, []);
```

src/screen/subIdentities/RegisterSubIdentitiesScreen.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  const onSetSubIdentitiesPress = async () => {
+    setSubmitSubsDisabled(true);
+    startTx({
+      address,
+      txMethod: 'identity.setSubs',
+      params: [subIdentities.map((sub) => [sub.accountId, {raw: sub.name}])],
+    })
+      .then(() => {
+        queryClient.invalidateQueries(['sub-identities', address]);
```
ahoseinian: I think the query key is different in the hook, how about
```suggestion
        queryClient.invalidateQueries(['sub-identities']);
```
it should clear everything starting with 'sub-identities'

src/screen/subIdentities/RegisterSubIdentitiesScreen.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  const onRemovePress = (accountId: string) => {
+    Alert.alert(
+      'Remove sub-identity',
+      `Do you want to remove account: \n ${accountId} ?`,
+      [
+        {
+          text: 'Yes',
+          onPress: () => {
+            setSubIdentities((prevInfos) => {
```
ahoseinian, line 84: Shouldn't we send a kind of transaction to remove a sub identity. I think it will just show up again when the cache gets cleared otherwise.

src/screen/subIdentities/RegisterSubIdentitiesScreen.tsx
silva-fj: When setting sub identities (making the actual transaction) we send all sub accounts (including the accounts already registered), so we need to keep them in a local state.. When adding and removing sub identities we are just updating the local state so later we can send all accounts in the transaction.

src/screen/subIdentities/RegisterSubIdentitiesScreen.tsx
silva-fj: Yeah I forgot to update this one after updating the queryKey in the hook.. I think `['sub-identities', {address}]` would also work to invalidate only the cache for the specific address

src/screen/subIdentities/RegisterSubIdentitiesScreen.tsx
silva-fj: We remove the sub-identity from the local state, the transaction is made with "Set sub-identities"

src/screen/subIdentities/RegisterSubIdentitiesScreen.tsx
ahoseinian: got it now ✅

src/screen/subIdentities/RegisterSubIdentitiesScreen.tsx
ahoseinian: so no second argument for useEffect here?

ahoseinian (approved):

PR #341, opened 2021-09-02
Refactoring Tx result

### LINKED ISSUE #331: Bug: What happens when last step of "set identity" fail ?
> As a user I'd like to understand why the submission of my identity doesn't work

After I've been through all different steps of "set identity", the app is stuck on this screen (waited for more than 5 minutes nothing happens) :
 

Finally I just touched another area of the screen which closes the and I end up on that screen where I have to begin the whole process again:
 

We have to understand what's happening there ? If it's a problem with the submission we should display an error message to the user so he doesn't wait hoping for something to happen. It's even better if we can tell the user what went wrong with the submission

silva-fj:
Getting the transaction status and errors directly from the transaction's result instead of looping over the events. In addition a warning step was added for the rare cases where a transaction was sent and the api got disconnected to give feedback to the  users. It will display a dialog similar to the error dialogs as shown down below

Screenshot_20210902-110330_Litentry

### DIFF
--- a/src/context/TxContext/index.tsx
+++ b/src/context/TxContext/index.tsx
@@ -1,21 +1,23 @@
+import React, {createContext, useCallback, useEffect, useMemo, useReducer, useRef} from 'react';
+import {Dimensions, StyleSheet} from 'react-native';
+import {Modalize} from 'react-native-modalize';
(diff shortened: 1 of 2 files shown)


ahoseinian (approved):
LGTM

PR #361, opened 2021-09-06
Feature/353 fix set identity action

### LINKED ISSUE #353: Keep showing the "set identity" action
> As a user I'd like to  have the option to complete an identity for which I've already submitted information

image

ahoseinian:
fixes #353 

- Combine 3 different state of my identity into one page
- Allow user to change identity after setting it once
- Show sub accounts for all accounts


src/screen/MyIdentityScreen/index.tsx
```diff
@@ -1,38 +1,277 @@
...
+                    </Text>
+                  )}
+                />
+              </MenuGroup>
+            </>
+          ) : null}
+
+          <Padder scale={1} />
+          <Button onPress={() => modalRef.current?.open()} status="basic">
+            {identity?.hasIdentity ? 'Complete Identity' : 'Set Identity'}
```
silva-fj: I think the word complete may confuse the users, I'm still not completely sure I understand the idea behind "Complete". How about **Update Identity** or **Edit Identity**?

ahoseinian: make sense to me, I've updated it to `Update Identity` hope @stephanie-olp is also fine with it :)

silva-fj (approved):

PR #364, opened 2021-09-08
POC - Fetching tips from our own api

### LINKED ISSUE #347: Indexer

silva-fj:
This PR is a follow up on these notes: https://github.com/litentry/litentry-app/issues/347#issuecomment-915205916

This is a POC and it is not meant to be merged. The relevant changes are in `src/screen/tips/TipsTest.tsx` (All the code needed to display the list of tips is in that file.), this is an example of how much our code can be simplified in the frontend apps if we fetch the data from our api instead of directly from the chain.  

To test it you can clone [this repository](https://github.com/litentry/litentry-graph) and run the api locally.

https://user-images.githubusercontent.com/8374946/132515011-021da966-e695-4719-a00a-deaab7e580f5.mp4

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -577,7 +577,7 @@ SPEC CHECKSUMS:
   BVLinearGradient: e3aad03778a456d77928f594a649e96995f1c872
   DoubleConversion: cf9b38bf0b2d048436d9a82ad2abe1404f11e7de
   FBLazyVector: e686045572151edef46010a6f819ade377dfeb4b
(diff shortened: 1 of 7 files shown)


ahoseinian:
I got couple of concerns about not having a database that we read from
- Would this scale up, how would we handle multiple clients requesting an endpoint which triggers multiple api requests. I mean there could be some kind of limit on how much a single api promise instance can request.
- We need a way to make sure the api connection wouldn't die without recovery on the server
- Without getting the data from a database the only thing the server does is keeping the logic that would have been replicated in the front-ends. we could also move this logic into a npm package and share it between the apps.

PR #438, opened 2021-09-30
Adding parachains overview screen

### LINKED ISSUE #409: Overview (summary + list)
As a user I'd like to have more information about the parachains

- Wireframe
image

@franjs I've picked the leases to be displayed but we can pick something that you think would be more relevant for the user. Of course we will have all the information on the details screen of each Parachain but we should display the most relevant one here

silva-fj:
Screenshot 2021-09-30 at 2 58 41 PM

There are some components than can be extracted and reused, later after all the PR's are merged we can iterate and do some refactoring.


ahoseinian (dismissed):
LGTM

ahoseinian (approved):

PR #440, opened 2021-09-30
Feature/410 crowdloans

ahoseinian:
Adds crowdloan screen and related hooks


src/api/hooks/useBestNumber.ts
```diff
@@ -5,17 +5,22 @@ import {BlockNumber} from '@polkadot/types/interfaces';
...
-  const {data} = useApiQuery('api_derive_chain_bestNumber', (apiPromise) => apiPromise.derive.chain.bestNumber());
+  const {data} = useApiQuery(BEST_NUMBER_QUERY_KEY, (apiPromise) => apiPromise.derive.chain.bestNumber());
 
   useEffect(() => {
     let unsub: VoidFn | undefined;
 
     (async () => {
       unsub = await api?.derive.chain.bestNumber((blockNumber) => {
-        queryClient.setQueryData('api_derive_chain_bestNumber', blockNumber);
+        const currentBestNumber = queryClient.getQueryData<BlockNumber>(BEST_NUMBER_QUERY_KEY);
```
silva-fj, line 20: You can also pass function to update the cache that takes the oldData: `queryClient.setQueryData(queryKey, oldData => newData)`

src/api/hooks/useBestNumber.ts
```diff
@@ -5,17 +5,22 @@ import {BlockNumber} from '@polkadot/types/interfaces';
...
+        if (!currentBestNumber?.eq(blockNumber)) {
```
silva-fj, line 21: 👍🏼

src/api/hooks/useFunds.ts
```diff
@@ -0,0 +1,167 @@
...
+    ? a.isCapped
+      ? -1
+      : 1
+    : a.isEnded !== b.isEnded
+    ? a.isEnded
+      ? 1
+      : -1
+    : 0;
+}
+////////////////////////////////////////////////////////////////////////////////////////////
```
silva-fj: What is this?

src/api/hooks/useBestNumber.ts
ahoseinian: I am trying to use that but it seems like if you call the setQueryData even with returning previous data it will cause rerenders. I will just keep it like this so the function wouldn't get called at all for now.

src/api/hooks/useFunds.ts
ahoseinian: Old habit to separate the code from types. I've removed it

silva-fj (dismissed):

ishrak-imam (dismissed):

silva-fj (approved):

PR #446, opened 2021-10-01
Bugfix/441 - Update data on network switch

ishrak-imam:
### Description
Appended the network name with the query key.


src/context/ChainApiContext.tsx
```diff
@@ -30,7 +29,7 @@ export function ChainApiContextProvider({children}: {children: React.ReactNode})
       return;
     }
 
-    queryClient.clear();
+    // queryClient.clear();
```
ishrak-imam: We may not have to clear the cache on network switch now. Also it results to a better UX.
@franjs @ahoseinian please comment what you guys think.

src/api/hooks/useApiQuery.ts
```diff
@@ -12,9 +12,10 @@ function useApiQuery<TData>(
   options: UseQueryOptions<TData> = {enabled: true},
 ) {
   const {api} = useApi();
+  const network = api?.runtimeChain != null ? api.runtimeChain.toString() : '';
 
   return useQuery(
-    queryKey,
+    `${queryKey}-${network}`,
```
ahoseinian: just one thing, the `queryKey` can be a string or array of string and objects. how about doing something like
```
[network, ...(typeof queryKey === 'string' ? [queryKey]: queryKey)]
```

src/context/ChainApiContext.tsx
ahoseinian: yes I guess it would be unnecessary.

src/context/ChainApiContext.tsx
```diff
@@ -30,7 +29,6 @@ export function ChainApiContextProvider({children}: {children: React.ReactNode})
       return;
     }
 
-    queryClient.clear();
```
silva-fj, line 32: `queryClient` is not used for anything else, can you remove the usage of `useQueryClient` as well?

src/api/hooks/useApiQuery.ts
```diff
@@ -12,16 +12,22 @@ function useApiQuery<TData>(
   options: UseQueryOptions<TData> = {enabled: true},
 ) {
   const {api} = useApi();
+  const network = api?.runtimeChain != null ? api.runtimeChain.toString() : undefined;
```
silva-fj, line 15: You can also do: 
  `const {currentNetwork} = useContext(NetworkContext);
`

silva-fj (approved):

src/api/hooks/useApiQuery.ts
ishrak-imam: `const {currentNetwork} = useContext(NetworkContext)`
I gave it a shot, but for some reason it does not behave as expected(refetch queries) for the screens in the drawer stack.

@franjs 
For now I am keeping it as it is. If you have time please ping me to discuss.

silva-fj: I'm not sure I follow, when we change the network we are actually changing the state in `NetworkContext` so it must be different, the issue is something else

PR #478, opened 2021-10-08
Feature/469 - Create account

ishrak-imam:
### Description
This is how the create account view looks so far.

The PR is still in `WIP`. This is just to take a look for you guys.


package.json
```diff
@@ -25,7 +25,9 @@
     "@eva-design/eva": "^2.1.1",
     "@polkadot/api": "^6.1.2",
     "@polkadot/apps-config": "^0.96.1",
+    "@polkadot/keyring": "^7.4.1",
```
silva-fj: There is no need to install these packages, they are already available.

silva-fj (commented):
It looks good, can you also add validation for the password? If it is a good password and that they match

src/screen/CreateAccountScreen.tsx
```diff
@@ -0,0 +1,109 @@
...
+import Clipboard from '@react-native-community/clipboard';
+
+type Account = {
+  title: string;
+  password: string;
+};
+
+export function CreateAccountScreen() {
+  const theme = useTheme();
+  const keyring = new Keyring();
```
ahoseinian: I think we need `ss58Format` for distinguishing between networks. https://polkadot.js.org/docs/keyring/start/ss58/

I think polkadot account addresses are supposed to start with 1 at beginning.
```suggestion
  const keyring = new Keyring({type: 'sr25519', ss58Format: 0});
```

ishrak-imam: We can add the `ss58Format` value in the networkContext for respective networks.

src/presentational/ProgressBar.tsx
```diff
@@ -5,32 +5,43 @@ import {StyleSheet, View} from 'react-native';
...
       </View>
       <View style={[styles.requiredContainer, {left}]} onLayout={(e) => setRequiredTextW(e.nativeEvent.layout.width)}>
-        <Text status="primary">{`${requiredAmount}% required`}</Text>
-        <View style={styles.requiredLine} />
+        {requiredAmount && (
+          <>
+            <Text status="primary">{`${requiredAmount}% required`}</Text>
+            <View style={styles.requiredLine} />
+          </>
+        )}
```
ishrak-imam: Needed to adjust the progress bar to reuse it as a password strength meter.

src/screen/CreateAccountScreen.tsx
ishrak-imam: Also I am not completely sure if we need to be concerned about using type `sr25519` because of the following:
https://polkadot.js.org/docs/keyring/start/create/#revisiting-crypto

src/screen/CreateAccountScreen.tsx
```diff
@@ -0,0 +1,150 @@
...
+export function CreateAccountScreen() {
+  const theme = useTheme();
+  const {currentNetwork} = React.useContext(NetworkContext);
+  const keyring = new Keyring({ss58Format: currentNetwork.ss58Format});
+  const [mnemonic] = React.useState(mnemonicGenerate());
+  const [account, setAccount] = React.useState<Account>({title: '', password: ''});
+  const [confirmPassword, setConfirmPassword] = React.useState('');
+  const [isMnemonicCopied, setIsMnemonicCopied] = React.useState(false);
+  const [isPasswordVisible, setIsPasswordVisible] = React.useState(false);
+  const [passwordStrength, setPasswordStrength] = React.useState(0);
```
ahoseinian: I think we can just calculate this no need to put this on state
```
const passwordStrength = zxcvbn(account.password).score
```

hanwencheng:
It would be good if we could support [the derivation path](https://wiki.polkadot.network/docs/learn-accounts#derivation-paths).

ishrak-imam:
> It would be good if we could support [the derivation path](https://wiki.polkadot.network/docs/learn-accounts#derivation-paths).

Thanks @hanwencheng 
Scope of this ticket is mainly UI focused. We are working in parallel and later on we will do sort of an integration with each other's work. So some aspects might seem incomplete for now.

About the `derivation path support`, it should be easily doable by extending our current work. We can prioritize that in the next iteration.

src/screen/CreateAccountScreen.tsx
ahoseinian: I've added the ss58Format in network context on this pr https://github.com/litentry/litentry-app/pull/481/files

src/presentational/ProgressBar.tsx
ishrak-imam: Changes are reverted back.

src/context/NetworkContext.tsx
```diff
@@ -30,6 +32,7 @@ const LitentryNetworkTest: NetworkType = {
   ws: ['wss://staging.registrar.litentry.io'],
   isTestnet: true,
   color: '#006400',
+  ss58Format: 9999, // don't know yet
```
silva-fj: It's `31`, you can find it [here](https://github.com/paritytech/substrate/blob/master/ss58-registry.json)

ahoseinian (dismissed):

silva-fj (dismissed):

ahoseinian (approved):

PR #479, opened 2021-10-08
Adding service to encrypt data with password

### LINKED ISSUE #468: Create a service to encrypt and decrypt

silva-fj:
I had some issues trying to use `crypto` in our project so I installed `react-native-aes-crypto`. The new package uses native libraries to implement AES encryption.

### DIFF
--- a/src/service/Encryptor.ts
+++ b/src/service/Encryptor.ts
@@ -1,0 +1,57 @@
+import Aes from 'react-native-aes-crypto';
+
+interface EncryptedData {
+  cipher: string;
+  iv: string;
+  salt: string;
+}
+
+async function generateSalt() {
+  return Aes.randomKey(32);
+}
+
+async function generateKeyFromPassword(password: string, salt: string) {
+  return Aes.pbkdf2(password, salt, 5000, 256);
+}
+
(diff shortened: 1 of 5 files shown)


ishrak-imam (approved):

ahoseinian (approved):
LGTM

src/service/Encryptor.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Encrypts an object using a password (AES encryption with native libraries)
+ *
+ * @example
+ * const encryptedData = await encrypt('secret', { data: 'data' });
+ *
+ */
+export async function encrypt(password: string, data: Record<string, unknown>) {
+  const salt = await generateSalt();
```
geastwood, line 36: @franjs
if `salt` is random, would it be problematic when decrypting?
Would be great we have a small unit test on this.

silva-fj: @geastwood The `salt` is included in the encrypted data (a json string that contains an object with the `cipher`, `salt` and `iv`) so it's available when decrypting. 

The thing is that `react-native-aes-crypto` uses native code so we have to mock all methods (not testing the real thing)

PR #482, opened 2021-10-12
Adds custom store for polkadot's ui-keyring

### LINKED ISSUE #472: Research : Pick library to store the data

silva-fj:
MMKV is faster than AsyncStorage and also can be encrypted (We will do it later)

TODO: 

- [x] Check android build


ahoseinian (dismissed):

ishrak-imam (dismissed):

geastwood:
How about the android build? I see it's still unchecked in the issue.

silva-fj:
> How about the android build? I see it's still unchecked in the issue.

Yeah, I'm looking into it. There is a workaround we have to do until we upgrade react-native to 0.66 because the library uses the new architecture (no bridge)

geastwood:
Ok, what currently stops us from upgrading to react-native 0.66?

silva-fj:
> Ok, what currently stops us from upgrading to react-native 0.66?

We need to wait until [react-native-reanimated 2.3.0](https://github.com/software-mansion/react-native-reanimated/releases) is released

geastwood:
> > Ok, what currently stops us from upgrading to react-native 0.66?
> 
> We need to wait until [react-native-reanimated 2.3.0](https://github.com/software-mansion/react-native-reanimated/releases) is released

Great, let's track this via issue. So we know the dependencies of upgrading to 0.66 and afterward what can we remove as temp hack (such as the android build).

silva-fj:
So the workaround does not work anymore, the latest versions of the library only work with `react-native` 0.66 so for now we can use `react-native-mmkv` 1.3.2 and we can upgrade after upgrading react-native

ishrak-imam (dismissed):

ahoseinian (approved):

PR #488, opened 2021-10-13
Adding secure keychain to store user's password

### LINKED ISSUE #487: Implement secure keychain
Implement secure Keychain to store user's password to encrypt data

silva-fj:
This PR adds a wrapper for Keychain (`react-native-keychain`) adding an extra layer of encryption before writing into the device's keychain

### DIFF
--- a/src/service/SecureKeychain.ts
+++ b/src/service/SecureKeychain.ts
@@ -1,0 +1,107 @@
+import {Platform} from 'react-native';
+import * as Keychain from 'react-native-keychain';
+import {encrypt, decrypt} from 'src/service/Encryptor';
+import {mmkvStorage} from 'src/service/MMKVStorage';
+
+const SALT = 'secret_phrase';
+const BIOMETRY_CHOICE = 'biometry_choice';
+const BIOMETRY_CHOICE_DISABLED = 'biometry_choice_disabled';
+const PASSCODE_CHOICE = 'passcode_choice';
+const PASSCODE_DISABLED = 'passcode_choice_disabled';
+const TYPES = {
+  BIOMETRICS: 'BIOMETRICS',
+  PASSCODE: 'PASSCODE',
+  REMEMBER_ME: 'REMEMBER_ME',
+};
+
+const defaultOptions = {
+  service: 'com.litentry',
+  authenticationPromptTitle: 'Authentication required',
+  authenticationPrompt: {title: 'Please authenticate in order to use Litentry App'},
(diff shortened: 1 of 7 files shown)


src/service/SecureKeychain.ts
```diff
@@ -0,0 +1,107 @@
+import {Platform} from 'react-native';
+import * as Keychain from 'react-native-keychain';
+import {encrypt, decrypt} from 'src/service/Encryptor';
+import {mmkvStorage} from 'src/service/MMKVStorage';
+
+const SALT = 'secret_phrase';
```
silva-fj: We could probably get this from a build variable, but this is used just as a phrase to "encode" the password to add an extra layer of security. They keychain is already encrypting the data.

src/service/SecureKeychain.ts
```diff
@@ -0,0 +1,107 @@
...
+  } else {
+    // Setting a password without a type does not save it
+    return await resetGenericPassword();
+  }
+
+  const encryptedPassword = await encryptPassword(password);
+  await Keychain.setGenericPassword('litentry-user', encryptedPassword, {...defaultOptions, ...authOptions});
+
+  if (type === TYPES.BIOMETRICS) {
+    mmkvStorage.set(BIOMETRY_CHOICE, true);
```
ahoseinian: Can we here do something like
```
mmkvStorage.set('selected_access_control', 'BIOMETRY' | 'PASSCODE');
```
instead of four different operations here and also on other lines below?

---
Also, I don't see this values being used somewhere. Can you describe how are we gonna use them?

src/service/SecureKeychain.ts
ahoseinian: Do you mean using env vars on build to inject the salt? I think that would also accessible as they explain [in react native docs](https://reactnative.dev/docs/security#storing-sensitive-info).

src/service/SecureKeychain.ts
silva-fj: Yep, we can also do that. Good idea

Later we will export them when writing the business logic for setting the password and so on.

src/service/SecureKeychain.ts
silva-fj: Yeah I know it will be accessible, this case is something we have to live with (unless we ask the user for a password to encrypt the password 😆 )

ahoseinian (approved):

ishrak-imam (approved):

PR #489, opened 2021-10-14
Add internal accounts in context

ishrak-imam:
### Description
Add internal accounts (created from seed/imported) into the AccountContext.

## Converted to draft. Please don't review now. Things will change....

### DIFF
--- a/src/context/ChainApiContext.tsx
+++ b/src/context/ChainApiContext.tsx
@@ -4,6 +4,7 @@ import {createLogger} from 'src/utils';
 import {NetworkContext} from './NetworkContext';
 import {keyring} from '@polkadot/ui-keyring';
 import {keyringStore} from 'src/service/KeyringStore';
+import {NetworkType} from 'src/types';
 
 const initialState: ChainApiContext = {
   status: 'unknown',
@@ -11,6 +12,8 @@ const initialState: ChainApiContext = {
   api: undefined,
   wsConnectionIndex: 0,
 };
+
+keyring.loadAll({store: keyringStore, type: 'sr25519'});
 
 export const ChainApiContext = createContext<ChainApiContext>(initialState);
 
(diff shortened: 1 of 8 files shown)


src/context/ChainApiContext.tsx
```diff
@@ -12,6 +13,8 @@ const initialState: ChainApiContext = {
   wsConnectionIndex: 0,
 };
 
+keyring.loadAll({store: keyringStore, type: 'sr25519'});
```
ishrak-imam, line 16: We can not load the keyring every time on network change since this is not allowed. So it can be loaded globally and we can just change the `ss58Format` on network change.

But this also means that on RN hot reload there will be an error 🤔

ahoseinian: can we do this on an effect inside `ChainApiContextProvider`, maybe that would fix the problem with RN hot reload.

```
useEffect(() => {
  keyring.loadAll({store: keyringStore, type: 'sr25519'});
}, [])
```

src/screen/CreateAccountScreen.tsx
```diff
@@ -39,7 +45,13 @@ export function CreateAccountScreen({navigation}: {navigation: NavigationProp<Ac
   );
 
   const onSubmit = () => {
-    keyring.addUri(mnemonic, account.password, {name: account.title});
+    const {json: key} = keyring.addUri(mnemonic, account.password, {name: account.title});
```
ahoseinian, line 48: I think we need to keep everything in keyring instead of having the keyring accounts duplicated in AccountContext. keyring has a method called `addExternal` that we can use to store the external accounts in it.

ishrak-imam: So technically AccountsContext is not needed anymore?

ahoseinian: That is my idea, the only thing that is different is `isFavorite` property. but maybe keyring also allows us to add that in `meta` object of the pair.

ishrak-imam:
Closing this PR with reference to https://github.com/litentry/litentry-app/pull/499

PR #499, opened 2021-10-15
Feature-470 Show internal accounts

ishrak-imam:
### Description
Showing internal accounts in Accounts screen.

Note:
Now every accounts action (add/import/import-json/remove/toggle favourite) is done through the ui-keyring.

So, the AccountsContext doesn't dispatch any accounts related action now. Alternatively it subscribe to the changes in the keyring and sets the accounts context for any change in the keyring.

### DIFF
--- a/src/screen/AccountsScreen.tsx
+++ b/src/screen/AccountsScreen.tsx
@@ -1,5 +1,4 @@
 import Identicon from '@polkadot/reactnative-identicon';
-import {keyring} from '@polkadot/ui-keyring';
 import {NavigationProp} from '@react-navigation/native';
 import {Button, Divider, Icon, ListItem, MenuItem, OverflowMenu, Text, useTheme} from '@ui-kitten/components';
 import {Account, useAccounts} from 'context/AccountsContext';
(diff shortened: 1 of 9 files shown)


src/screen/AccountsScreen.tsx
```diff
@@ -35,7 +37,10 @@ export function AccountsScreen({navigation}: {navigation: NavigationProp<Complet
   const sortByFunction = sortBy === 'name' ? sortByDisplayName : sortByIsFavorite;
   const [sortMenuVisible, setSortMenuVisible] = React.useState(false);
 
-  const internalAccounts = keyring.getAccounts();
+  const onToggleFavorite = (address: string) => {
+    const pair = keyring.getPair(address);
```
silva-fj, line 41: What if the account is external? Do we still get a pair after using `keyring.addExternal`?

src/screen/DevScreen.tsx
```diff
@@ -87,7 +87,7 @@ function DevScreen() {
             <Divider />
             {accounts.map((account) => (
               <View key={account.address}>
-                <ListItem
+                {/* <ListItem
```
silva-fj, line 90: What's up with this? 😆

silva-fj (approved):

src/screen/AccountsScreen.tsx
ishrak-imam: Yes, we still get a pair but may be signing any transaction is not possible with that without using any external signer.

src/screen/DevScreen.tsx
ishrak-imam: Some dev features, we might not need anymore 😆

PR #504, opened 2021-10-18
Adding flow to sign transactions for internal accounts

### LINKED ISSUE #473: Signing transaction with internal account

### LINKED ISSUE #493: Ask for password for signing transactions
Every time the user wants to sign a transaction with the internal wallet, we should ask him for a password to enable that action

silva-fj:
It works exactly the same as for external accounts (how we currently sign transactions) but without the screens for scanning the QR codes. 
 
If the internal account is locked, the user has to enter the password in a form.

Screenshot 2021-10-19 at 11 22 44 AM

Screenshot 2021-10-19 at 11 23 49 AM

For internal accounts, pressing continue after the preview will send the transaction showing the submitting view as usual.

Screenshot 2021-10-19 at 12 23 43 PM

### DIFF
--- a/src/context/TxContext/index.tsx
+++ b/src/context/TxContext/index.tsx
@@ -3,6 +3,7 @@ import {Dimensions, StyleSheet} from 'react-native';
 import {Modalize} from 'react-native-modalize';
 import QRCodeScanner from 'react-native-qrcode-scanner';
 import {ApiPromise} from '@polkadot/api';
+import {keyring} from '@polkadot/ui-keyring';
 import {SubmittableExtrinsic} from '@polkadot/api/submittable/types';
 import {SignerPayloadJSON, SignerResult} from '@polkadot/types/types';
 import {BN_ZERO} from '@polkadot/util';
@@ -18,7 +19,6 @@ import {formatCallMeta} from 'src/packages/call_inspector/CallInspector';
 import AsyncSigner from 'src/service/AsyncSigner';
 import globalStyles, {standardPadding} from 'src/styles';
 import WarningDialog from 'presentational/WarningDialog';
-import {getPair} from 'src/utils';
 
 let id = 0;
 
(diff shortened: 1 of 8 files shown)


src/context/TxContext/index.tsx
```diff
@@ -76,28 +82,61 @@ function TxContextProvider({children}: PropTypes): React.ReactElement {
...
+      const info = await transaction.paymentInfo(address);
+      dispatch({
+        type: 'SHOW_TX_PREVIEW',
+        payload: {
+          txPayload: txPayload,
+          params,
+          title,
+          description,
+          partialFee: info.partialFee.toNumber(),
+          isExternalAccount: keyringPair === undefined,
```
silva-fj: @ishrak-imam is this still true after your changes? Or we will get a `pair` also for external addresses?

silva-fj: oh, the changes are already merged.. I'll rebase and take a look

ishrak-imam (dismissed):
So now only thing left is, prompt for the password/pick it up from keychain using biometric confirmation and use that to unlock the pair before signing?

silva-fj:
> So now only thing left is, prompt for the password/pick it up from keychain using biometric confirmation and use that to unlock the pair before signing?

I have added the view to unlock the account and as discuss I will create another issue to allow the user store the password in the keychain so they can use biometrics to retrieve it

ishrak-imam (approved):
So, the keyring pair is a reference. Once it is unlocked we can just use it to sign.

Should we also lock the pair `keyringPair.lock()` once the signing is done?

ahoseinian (approved):

silva-fj:
> So, the keyring pair is a reference. Once it is unlocked we can just use it to sign.
> 
> Should we also lock the pair `keyringPair.lock()` once the signing is done?

I don't think we need to lock it again, the user will have the preview anyways before making the transaction and when closing the app the keyring will lock the account again

PR #506, opened 2021-10-19
Fix api being set twice at start of the app

ahoseinian:
### Effect on `ChainApiContext` called twice
This was caused by changes in network context. `usePersistedState` was using the default value and after a millisecond was changing it with the value stored in async storage. technically it shouldn't have much effect but after fixing it I was not getting the connection issues to api anymore and the app stays responsive. although not very fast still!

I have tested removing `mmkv` to test the speed with chrome debugging enabled. it was a bit faster. but still let's wait for the next version of RN.

CreateAccountScreen is specially slow. I think some kind of performance problem is going on there. we need to check it out. @ishrak-imam


src/context/NetworkContext.tsx
```diff
@@ -55,21 +55,15 @@ type PropTypes = {
...
-      currentNetwork,
-      availableNetworks,
-      select,
-    }),
-    [currentNetwork, select],
+    () => (currentNetwork ? {currentNetwork, availableNetworks, select: setCurrentNetwork} : undefined),
+    [currentNetwork, setCurrentNetwork],
   );
 
+  // Wait for async storage response
```
silva-fj: Did you try using replacing `AsyncStorage` for `MMKV` in `usePersistedState`? It should be faster and also it's synchronous

src/hook/usePersistedState.ts
```diff
@@ -6,14 +6,11 @@ const logger = createLogger('usePersistedState');
 
 export type PersistedStateKey = 'network' | 'accounts' | 'theme' | 'selected_push_topics';
 
-export function usePersistedState<T>(key: PersistedStateKey): [state: T | undefined, setState: (data: T) => void];
-export function usePersistedState<T>(key: PersistedStateKey, initialState: T): [state: T, setState: (data: T) => void];
-
 export function usePersistedState<T>(
   key: PersistedStateKey,
-  initialState?: T,
+  fallback?: T,
```
silva-fj: I'm not sure I understand the changes here. We need the ability to set an initial state, this is not the same as always returning a fallback value since the initial state is only set once, the first time the hook runs. Also the fallback value will trigger an extra re-render when calling `setState(fallback)`, If one needs a fallback it would be better to do it in the component e.g:

```
const [value] = usePersistedState('key');
const x = Boolean(value) ? value : fallback;
```

ahoseinian: in this case it would behave like this
- first render: `value` is `undefined` because async storage is async and it takes time to get data. x would be `fallback`.
- second render: value has come from async storage. x would change from `fallback` to `value`. 
This causes the hook in sub component to run once with `fallback` and immediately afterwards with `value`

---- 

I tried to make it so
- on first render: `value` would be `undefined` and we would not render any subcomponents for `undefined` value
- second render:  async storage result has come back. if result exists value = result if not value = fallback.

therefor sub comps would only receive either `fallback` or `value`. not `fallback` and then `value`

src/context/NetworkContext.tsx
ahoseinian: no I didn't check that out! I will try. thanks

ahoseinian:
@franjs @ishrak-imam on the device the app crashes because api starts to throw errors. doesn't happen on emulator. any guess into what can be the cause? I am out of ideas atm.
Screenshot 2021-10-19 at 12 04 28

ishrak-imam:
> on the device the app crashes because api starts to throw errors. doesn't happen on emulator.

The same error also happens if hermes is enabled in emulator.

silva-fj:
@ahoseinian is this an android only issue?

src/hook/usePersistedState.ts
silva-fj: I understand what you mean, we would be optimizing for this specific use case. The idea of this hook is to have the same api (without lazy initialization) as `useState`.. The only issue is the async call, I hope we could solve it by using MMKV.

ahoseinian:
> @ahoseinian is this an android only issue?

Yes just happens on android devices. android emulator & iOS are fine

src/hook/usePersistedState.ts
ahoseinian: I turned this to previous version and used mmkv instead to make it sync 👍

src/hook/useIsFirstAppStart.ts
```diff
@@ -1,5 +1,5 @@
 import {useQuery} from 'react-query';
-import * as AsyncStorage from 'src/service/AsyncStorage';
+import * as AsyncStorage from 'src/service/PersistedObjectStorage';
```
silva-fj:
```suggestion
import * as Storage from 'src/service/PersistedObjectStorage';
```

src/hook/usePushTopics.ts
```diff
@@ -1,5 +1,5 @@
 import {useQueryClient, useQuery, useMutation} from 'react-query';
-import * as AsyncStorage from 'src/service/AsyncStorage';
+import * as AsyncStorage from 'src/service/PersistedObjectStorage';
```
silva-fj:
```suggestion
import * as Storage from 'src/service/PersistedObjectStorage';
```

silva-fj (dismissed):

silva-fj (approved):

PR #510, opened 2021-10-19
Upgrading react-native to 0.66.1

silva-fj:
Also `react-native-reanimated` and `react-native-mmkv` needed to be upgraded.

After this PR is merged, make sure to clean the build/cache if you run into any issues


ahoseinian (approved):

ishrak-imam (approved):
We are not enabling hermes on iOS?

silva-fj:
> We are not enabling hermes on iOS?

We can give it a try later, I don't know how stable it is on IOS

PR #527, opened 2021-10-26
Add leases to parathreads screen

### LINKED ISSUE #461: Add lease information in parathread list item

ahoseinian:
Fixes #461
I couldn't test it with live data because currently all parathreads have `None` as leases but the logic should be fine and it will show like the image below if they would have the leases.


ishrak-imam (approved):
LGTM

silva-fj (approved):

PR #532, opened 2021-10-27
Feature/connection retry

ishrak-imam:
### Description

Added connection retry/change network screen on `disconnect` and `error` event.

https://user-images.githubusercontent.com/16683923/139058216-4404c6ee-5ade-4198-be72-3f779bec8243.mp4

### DIFF
--- a/src/context/ChainApiContext.tsx
+++ b/src/context/ChainApiContext.tsx
@@ -4,6 +4,8 @@ import {createLogger} from 'src/utils';
 import {NetworkContext} from './NetworkContext';
 import {keyring} from '@polkadot/ui-keyring';
 import {keyringStore} from 'src/service/KeyringStore';
+import {useNavigation} from '@react-navigation/core';
+import {connectionRetryScreen} from 'src/navigation/routeKeys';
 
 const initialState: ChainApiContext = {
   status: 'unknown',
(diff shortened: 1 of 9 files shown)


src/screen/ConnectionRetryScreen.tsx
```diff
@@ -0,0 +1,85 @@
...
+    StackNavigationProp<RootStackParamList>
+  >;
+};
+
+export function ConnectionRetryScreen({navigation}: PropTypes) {
+  const {currentNetwork, select} = useContext(NetworkContext);
+  const {status} = useContext(ChainApiContext);
+
+  const onRetry = () => {
+    select({...currentNetwork}); // to make the useEffect hook run again in ChainApiContext
```
silva-fj: How about exposing a function from `ChainApiContext` to reconnect?

ishrak-imam: Thought about that, but I got lazy :p 
I'll do that, was aiming for collecting some feedbacks for now.

src/context/ChainApiContext.tsx
```diff
@@ -46,17 +55,27 @@ export function ChainApiContextProvider({children}: {children: React.ReactNode})
     function handleReady() {
       logger.debug('ChainApiContext: Api ready at', wsAddress);
       keyring.setSS58Format(currentNetwork.ss58Format);
-      dispatch({type: 'ON_READY', payload: apiPromise});
+      dispatch({
+        type: 'ON_READY',
+        payload: {
+          apiPromise,
+          reconnect: (network: NetworkType) => {
```
ahoseinian: wouldn't reconnect be always called with the currentNetwork? is this necessary to pass the network again? also I don't know about saving this function in state. can't we just add in to provider value?

silva-fj (approved):

PR #537, opened 2021-10-27
Feature/525 crowdloan auction contribution

### LINKED ISSUE #525: Feature: Crowdloan auction contribution
> As a dot holder I wanna contribute to a crowdloan 

More details : We wanna let the users participate in an active crowdloan within the app

On the crowdloan screen (Parachain section), user should see a "contribute" button action next to each ongoing crowdloan:

! : the Litentry crowdloan teaser you can see on the wireframe will be tackled in a separate ticket (add link) 

When user clicks on one of those buttons he should see the following pop up:

**Some links shared by Ben:** 
The extrinsic:
(https://github.com/litentry/applications/blob/main/crowdloan/utils/contribute.ts)
A dummy page that just fires it on load:
https://github.com/litentry/applications/blob/main/crowdloan/pages/test-contribute.tsx
And this is the hook to get the metadata about the crowdloan (it was a real pain in the arse):
(https://github.com/litentry/applications/blob/main/packages/substrate-hooks/src/useCrowdloanData.ts)

ahoseinian:
Add ability to contribute for crowd loans

I couldn't test the success case as it needs some KSM to be contributed.

fixes #525

https://user-images.githubusercontent.com/1168983/139091412-48e7b48d-dd89-4360-bcd1-d55d36ccd4b4.mp4

### DIFF
--- a/src/presentational/SelectAccount.tsx
+++ b/src/presentational/SelectAccount.tsx
@@ -1,14 +1,14 @@
 import {IndexPath, Select, SelectItem} from '@ui-kitten/components';
-import {useAccounts} from 'context/AccountsContext';
+import {Account} from 'context/AccountsContext';
 import React from 'react';
 
 interface Props {
+  accounts: Account[];
   selected?: string;
   onSelect: (address: string) => void;
 }
 
-export function SelectAccount({selected, onSelect}: Props) {
-  const {accounts} = useAccounts();
+export function SelectAccount({selected, onSelect, accounts}: Props) {
   const selectedIndex = accounts.findIndex((a) => a.address === selected);
   const selectedIndexPath = selectedIndex > -1 ? new IndexPath(selectedIndex) : undefined;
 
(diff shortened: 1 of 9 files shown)


ahoseinian:
I noticed that on device interactions with keyring causes a lot of delay and make the app unusable. needs to be addressed in separate ticket

src/presentational/SelectAccount.tsx
```diff
@@ -1,14 +1,14 @@
...
 
 interface Props {
+  accounts: Account[];
   selected?: string;
   onSelect: (address: string) => void;
 }
 
-export function SelectAccount({selected, onSelect}: Props) {
-  const {accounts} = useAccounts();
+export function SelectAccount({selected, onSelect, accounts}: Props) {
```
silva-fj, line 11: Why not getting the accounts from the hook?

ahoseinian: for some reason the context was not field inside of the `Modal` component that we are using. so I thought it's better to pass it from outside. don't know why previously it wasn't an issue

silva-fj (approved):

PR #551, opened 2021-11-04
Feature/substrate sign

ahoseinian:
I have prepared the PR for substrate-sign here. the persistent storage is also included.

as mentioned before the JSON import/export are not compatible with polkadot.js.org but it would work for exporting accounts to different devices using our app.


src/context/AccountsContext.tsx
```diff
@@ -1,84 +1,94 @@
...
-  }
+function AccountsProvider({children}: {children: React.ReactNode}) {
+  const [accounts, setAccounts] = usePersistedState<Accounts>('accounts', {});
+
+  const value = {
+    accounts,
+    addAccount: (account: Account) => {
+      setAccounts({...accounts, [account.address]: account});
+    },
+    setAccountFavorite: (address: string, isFavorite: boolean) => {
```
silva-fj: how about `toggleFavorite(address: string) => void` and change the value here instead of passing `isFavorite` from the callers

silva-fj (dismissed):

src/context/AccountsContext.tsx
ahoseinian: done ✅

silva-fj (approved):

PR #553, opened 2021-11-05
Keyring webview

ishrak-imam:
### Description

Basic implementation of creating a keyring account using the packages through webview so that we can leverage the WASM support.

## TODO:

Feature:
- [ ] Add signing transactions support

Improvements:
- [ ] Fix TS errors
- [ ] Encapsulate setCallback within AccountsContext
- [ ] Try to avoid setTimeout while html is loading in webview
- [ ] Use the dispatched actions type strings from same file in both projects
- [ ] Add TS and linting support for the litentry-keyring project


geastwood:
Does this satisfy all our need in them of keyring functions?

ishrak-imam:
> Does this satisfy all our need in them of keyring functions?

Communication between RN and webview is not straight forward, involves message passing in string format i.e. `JSON parse and stringify`. Although this implementation has much room for improvements such as automating the build process of webview html, improving the code quality etc. it supports all our current needs as in, creating internal/external accounts, importing account from seed or json file.

Also one of the other non trivial part of this approach is that webview local storage will not persist data in between app starts. So we have to store all the accounts data in RN app local store and populate the webview local storage on app start so that webview keyring has access to the previously created accounts

geastwood:
And how's the performance?

ahoseinian:
Maybe we can use https://www.npmjs.com/package/json-rpc-2.0 to encapsulate the communication between WebViews and RN.

ishrak-imam:
> And how's the performance?

Better than the normal approach, i.e. the app would be usable at least.

src/context/AccountsContext.tsx
```diff
@@ -14,67 +18,270 @@ export type Account = {
...
+    if(localStoreAccounts.length > 0) {
+      mmkvStorage.set(KEYRING_ACCOUNTS, JSON.stringify(localStoreAccounts))
+    }
+  }, [localStoreAccounts])
 
-  return <AccountsContext.Provider value={accounts}>{children}</AccountsContext.Provider>;
-}
+  useEffect(() => {
+    if(html) {
+      setTimeout(() => {
```
silva-fj, line 133: Why do we need `setTimeout`?

src/screen/CreateAccount/CreateAccountScreen.tsx
```diff
@@ -26,6 +26,20 @@ export function CreateAccountScreen({
 
   const theme = useTheme();
   const {currentNetwork} = React.useContext(NetworkContext);
+  const {setCallback, createAccount, addAccount} = React.useContext(AccountsContext)
+
+  setCallback((data) => {
```
silva-fj, line 31: This should be encapsulated within the `AccountsContext`.

silva-fj (commented):
Great work @ishrak-imam, let's keep this branch around in case we need to fallback to this solution later.

You can add a TODO to add the ability to sign transactions getting the keyringPair from the keyring and also add comments on the things you know we would need to improve so we can remember

src/context/AccountsContext.tsx
ishrak-imam: It takes some time to load the html into the webview, before that the webviewref is not ready.

Could have been better if it is possible to listen for the ref change.

Issue #552, opened 2021-11-04
Bug: Tips aren't displayed properly


stephanie-olp:
App version: Build 211 on iOS

On both Polkadot and Kusama networks, the list of tips isn't displayed properly. We can only see one tip while on Polkadot.js there are many 
![image](https://user-images.githubusercontent.com/85941267/140374661-ea89326c-76c3-4bbd-b6e7-b18918da8581.png)

stephanie-olp:
I've just downloaded the latest version and the problem seems to be fixed :) 
![image](https://user-images.githubusercontent.com/85941267/140940833-c24dc631-4070-46fd-a018-b3121b64df16.png)

PR #562, opened 2021-11-09
Bug 552- Tips aren't displayed properly

ishrak-imam:
### Description

[Bug](https://github.com/litentry/litentry-app/issues/552)

The bug was not reproducible. However, added memoization for the TipTease component.


ahoseinian (dismissed):

silva-fj (dismissed):

silva-fj:
I can reproduce the issue, it takes a couple of seconds to load the data in the teasers, which is weird bacause the data should be already available since it is not displaying the loading view. 

@ishrak-imam can you add a loading screen for the tip detail screen?

ishrak-imam:
`useAccountIdentityInfo` and `useTipReason` these two hooks actually takes a bit of time to resolve the data, that's why added a loading view where these are used.

silva-fj (approved):

ahoseinian (approved):

PR #575, opened 2021-11-11
Bug-573 verify mnemonic

### LINKED ISSUE #573: Bug: New account creation can't verify mnemonic
App version: Latest

## Steps To Reproduce

1. Create new account
2. Will be presented with mnemonic words from the App, see screenshot
3. Since there is the same word in the mnemonic words, i.e. `leopard` appeared twice, can't pass the validation

IMG_3043

https://user-images.githubusercontent.com/1623359/141138708-2d5a3669-efd8-4ae9-93e8-7e3f2f999732.MP4

A possible solution will be simply to loop the generation of random mnemonic words, if there are any duplicates in the 12 words, generate again.

ishrak-imam:
### Description

Fixes https://github.com/litentry/litentry-app/issues/573


ahoseinian (dismissed):
I thought having the same text twice is a bug of SubstraiteSign but checking it on polkadot.js.org it also doesn't show a validation error so should be fine 👍 

Screenshot 2021-11-11 at 09 08 06

src/screen/CreateAccount/MnemonicScreen.tsx
```diff
@@ -9,12 +9,18 @@ import FormLabel from 'presentational/FormLabel';
...
+
+const logger = createLogger('Mnemonic');
 
 export function MnemonicScreen({navigation}: {navigation: NavigationProp<AccountsStackParamList>}) {
   const [mnemonic, setMnemonic] = React.useState<string>();
 
   React.useEffect(() => {
-    SubstrateSign.randomPhrase(12).then(setMnemonic);
+    SubstrateSign.randomPhrase(12).then((mnemonic) => {
+      logger.debug('Mnemonic seed ::: ', mnemonic);
```
silva-fj: let's remove this or only log in development

silva-fj (dismissed):

src/screen/CreateAccount/MnemonicScreen.tsx
ishrak-imam: This is to see the mnemonic so that we can verify while testing without writing it down somewhere or taking a screenshot. I assumed `logger.debug` was only logging in dev mode.

silva-fj (approved):

src/screen/CreateAccount/MnemonicScreen.tsx
silva-fj: Right now it's fine, but later we can integrate the logger with Sentry for example.. we just need to change the transport in the config and we may forget that we are logging the mnemonic. Debug is just the priority level so you can filter in this kind of tools

Issue #589, opened 2021-11-12
Feature: Reconnect to API automatically when app comes to foreground


ishrak-imam:
When the app is in background long enough it gets disconnected from the API.

It should try to reconnect automatically once it is in foreground.

PR #592, opened 2021-11-12
Feature-589 Automatic reconnect if app disconnects in background

ishrak-imam:
### Description

Feature: https://github.com/litentry/litentry-app/issues/589

### DIFF
--- a/src/hook/useAppBackgroundApiReconnect.ts
+++ b/src/hook/useAppBackgroundApiReconnect.ts
@@ -1,27 +1,27 @@
-import {useEffect} from 'react';
+import {useEffect, useRef} from 'react';
 import {useApiReconnect} from 'context/ChainApiContext';
 import {useNavigation} from '@react-navigation/core';
 import {apiLoadingScreen} from 'src/navigation/routeKeys';
-import {useAppState} from './useAppState';
+import {useAppState} from 'src/hook/useAppState';
 
 export function useAppBackgroundApiReconnect() {
   const {canReconnect, reconnect} = useApiReconnect();
   const navigation = useNavigation();
-  const {didAppCameToForeground, setDidAppCameToForeground} = useAppState();
+  const appState = useAppState();
+  const didAppCameToForegroundOnce = useRef(false);
 
   useEffect(() => {
-    if (didAppCameToForeground && canReconnect) {
-      /**
-       * make it false once used, otherwise it will try to reconnect on any
-       * type of disconnect event. Such as we don't want automatic reconnect
-       * when the app is already in foreground.
-       */
-      setDidAppCameToForeground(false);
+    didAppCameToForegroundOnce.current = appState.didAppCameToForeground;
+  }, [appState]);
 
+  useEffect(() => {
+    if (didAppCameToForegroundOnce.current && canReconnect) {
       reconnect();
       setTimeout(() => {
         navigation.navigate(apiLoadingScreen);
       }, 0);
+
+      didAppCameToForegroundOnce.current = false;
     }
-  }, [canReconnect, navigation, reconnect, didAppCameToForeground, setDidAppCameToForeground]);
+  }, [canReconnect, navigation, reconnect, appState]);
 }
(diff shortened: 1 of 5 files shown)


src/hook/useAppBackgroundApiReconnect.ts
```diff
@@ -0,0 +1,27 @@
...
+  const {didAppCameToForeground, setDidAppCameToForeground} = useAppState();
+
+  useEffect(() => {
+    if (didAppCameToForeground && canReconnect) {
+      /**
+       * make it false once used, otherwise it will try to reconnect on any
+       * type of disconnect event. Such as we don't want automatic reconnect
+       * when the app is already in foreground.
+       */
+      setDidAppCameToForeground(false);
```
ahoseinian: This should work but seems to me it would be easier if we remove the `useAppState` hook and put the App state listener here. because the other hook wouldn't be reusable in any other place when we change it's state here.

silva-fj: +1

ishrak-imam: Bringing in the AppState listener ~does not solve the problem~ creates a different problem. [Here](https://github.com/litentry/litentry-app/pull/592/commits/6996d7066451a66411088fd13cfbde78d6d6b122) is another fix.

@ahoseinian Thanks for the nice catch here though.

src/hook/useAppState.ts
```diff
@@ -0,0 +1,28 @@
...
+  const appStateRef = useRef(AppState.currentState);
+  const [appState, setAppState] = useState({
+    isActive: false,
+    didAppCameToForeground: false,
+  });
+
+  useEffect(() => {
+    const subscription = AppState.addEventListener('change', (nextAppState) => {
+      const isNextStateActive = nextAppState === 'active';
+      setAppState({...appState, isActive: true});
```
silva-fj:
```
setAppState({
    ...appState, 
    isActive: true,
    didAppCameToForeground: appStateRef.current.match(/inactive|background/) && isNextStateActive
});
```

src/hook/useAppState.ts
```diff
@@ -0,0 +1,28 @@
...
+      if (appStateRef.current.match(/inactive|background/) && isNextStateActive) {
+        setAppState({...appState, didAppCameToForeground: true});
+      }
+      appStateRef.current = nextAppState;
+    });
+
+    return () => {
+      subscription.remove();
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
silva-fj: Why not including the dependencies that are used inside the effect?

src/hook/useAppBackgroundApiReconnect.ts
```diff
@@ -0,0 +1,27 @@
...
+  const didAppCameToForegroundOnce = useRef(false);
+
+  useEffect(() => {
+    didAppCameToForegroundOnce.current = appState.didAppCameToForeground;
+  }, [appState]);
+
+  useEffect(() => {
+    if (didAppCameToForegroundOnce.current && canReconnect) {
+      reconnect();
+      setTimeout(() => {
```
silva-fj, line 20: why do we need this?

src/hook/useAppState.ts
ishrak-imam: Here the only dependency is `appState` and once the listener is registered this hook does not need to run anymore.

src/hook/useAppBackgroundApiReconnect.ts
ishrak-imam: Once the api disconnects the app tries to navigate to `connectionRetryScreen` from ChainApiContext and to `apiLoadingScreen` from here.
On android this back to back navigation doesn't work properly.

src/hook/useAppState.ts
ishrak-imam: It is not needed anymore. The hook does not have any dependency now.

silva-fj (approved):

PR #598, opened 2021-11-17
Feature/add rn paper

### LINKED ISSUE #498: Change the header
Allt the header's style are here:
https://www.figma.com/file/zadQiUwhMXr9632TBORG2P/Design-System---RN-App?node-id=723%3A4

ahoseinian:
Fixes #498 

- This would not implement the design header, we are just using the material header provided by the rn-paper library.
- All header components are gathered in one place, which should help changing the under lying ui package in future much easier
- RNPaper is added


ios/Litentry/Info.plist
```diff
@@ -45,13 +45,28 @@
...
 		<string>DMSans-BoldItalic.ttf</string>
 		<string>DMSans-Italic.ttf</string>
 		<string>DMSans-Medium.ttf</string>
 		<string>DMSans-MediumItalic.ttf</string>
 		<string>DMSans-Regular.ttf</string>
+		<string>AntDesign.ttf</string>
+		<string>Entypo.ttf</string>
+		<string>EvilIcons.ttf</string>
+		<string>Feather.ttf</string>
+		<string>FontAwesome.ttf</string>
```
silva-fj: We should probably include only the icons we are going to use (Material*Icons)

src/screen/ConnectionRetryScreen.tsx
```diff
@@ -34,14 +35,17 @@ export function ConnectionRetryScreen({navigation}: PropTypes) {
...
-      <ScreenNavigation
-        renderTitle={() => (
-          <View style={styles.networkName}>
-            <Text category="s1">Litentry</Text>
-            {currentNetwork ? <NetworkItem item={currentNetwork} isConnected={false} /> : null}
-          </View>
-        )}
-      />
+      <AppHeader>
+        <AppBar.Action icon="" tvParallaxProperties={undefined} hasTVPreferredFocus={false} onPress={noop} />
```
silva-fj: Do we need to pass all these props?

src/packages/base_components/AppBars.tsx
```diff
@@ -0,0 +1,62 @@
...
+  subtitle,
+}: {
+  onActionLeftPress: () => void;
+  onContentPress: () => void;
+  subtitle?: React.ReactNode;
+}) {
+  return (
+    <AppHeader>
+      <AppBar.Action
+        tvParallaxProperties={undefined}
```
silva-fj: We can probably just drop the prop since you are passing `undefined`

src/screen/ConnectionRetryScreen.tsx
ahoseinian: There is some sort of ts issue going on here. bad typings in repaper I suppose. it will error without this tv props.

src/packages/base_components/AppBars.tsx
ahoseinian: as mentioned it previous comment, would show an ts issue without it. couldn't find a better way to fix this.

src/screen/ConnectionRetryScreen.tsx
ahoseinian: sorry, it's already fixed https://github.com/callstack/react-native-paper/issues/2965

silva-fj (approved):

PR #610, opened 2021-11-19
Fix iOS status bar

ishrak-imam:
### Description

Change iOS status bar style based on theme.


src/packages/base_components/AppBars.tsx
```diff
@@ -29,6 +30,7 @@ export function MainStackAppBar({
 }
 
 export function AppHeader({children}: {children: React.ReactNode}) {
+  useIOSStatusBarStyle();
```
ahoseinian: the idea was to make this components here independent of hooks or any other app related logic. plus I think this way the hook would be mounted multiple times which is not necessary. could we instead use this hook in App.tsx or just put the logic in theme provider ?

silva-fj (approved):

PR #614, opened 2021-11-22
Feature/586 add auctions screen

### LINKED ISSUE #586: Auction Screen overview
Wireframe : 
image

ahoseinian:
I decided to go with the pagination approach because the list was too long:


src/screen/Parachains/ParachainAuctionsScreen.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const {items} = props;
+  const numberOfItemsPerPage = 8;
+
+  const from = page * numberOfItemsPerPage;
+  const to = Math.min((page + 1) * numberOfItemsPerPage, items.length);
+  const pageItems = props.items.slice(from, to);
+
+  React.useEffect(() => {
+    setPage(0);
+  }, [numberOfItemsPerPage]);
```
silva-fj: `numberOfItemsPerPage` will never change. Why do we need this effect?

ahoseinian: That is what happens when I blindly copy paste code :) removed it 👍

silva-fj (approved):

PR #625, opened 2021-11-24
Feature/515 accounts list UI

### LINKED ISSUE #515: Change Accounts list UX
As a user, I wanna visualize my accounts in a way I can distinguish my internal account from the external ones.

1.  **When the user doesn't have an internal account yet**, he should see the following action buttons under the "your internal account" section:  _Import seed_ & _Generate new seed_

Mockup: 
 

The button _Add external account_ should always be visible to the user under the "Other accounts (external)" section, whether he already has an account or not

2. **When the user already has an internal account **
Only the action **Add external account** should be visible
Mockup:

ahoseinian:
fixes #515 

@stephanie-olp because the ability to have more than one external account exists the designs provided in the ticket would not work. as discussed in the last meeting about this, I have used Floating action buttons instead and change the design as in the provided pictures. hope that is also fine?

new styles:


src/screen/AccountsScreen.tsx
```diff
@@ -189,28 +179,70 @@ function AccountItem({
...
+const Buttons = ({navigation}: {navigation: Props['navigation']}) => {
+  const [state, setState] = React.useState({open: false});
+
+  const onStateChange = ({open}: {open: boolean}) => setState({open});
+
+  const {open} = state;
+
+  return (
+    <Provider>
+      <Portal>
```
ishrak-imam: Does it work properly?
The app should be already wrapped with Provider in themeContext

ahoseinian: yes, without this provider the FAB.Group would show up over the whole app ( all navigation screens ) and with this Provider here it would be just inside this page.

ishrak-imam: Probably provider seems to be working like that but there's still a problem. It is taking up screen space.
The problem is visible if the list is long.

stephanie-olp:
> fixes #515
> 
> @stephanie-olp because the ability to have more than one external account exists the designs provided in the ticket would not work. as discussed in the last meeting about this, I have used Floating action buttons instead and change the design as in the provided pictures. hope that is also fine?
> 
> new styles:
> 
>   

Talked with Lilly and she likes the idea of floating buttons: :)

src/screen/AccountsScreen.tsx
ahoseinian: yes thanks for noticing, I wrapped the provider around the whole page and that fixed it.

src/screen/AccountsScreen.tsx
```diff
@@ -38,104 +58,77 @@ export function AccountsScreen({navigation}: {navigation: NavigationProp<Complet
...
-                  setSortMenuVisible(false);
-                  switch (row) {
-                    case 0:
-                      setSortBy('name');
-                      break;
-                    case 1:
-                      setSortBy('favorites');
-                      break;
-                  }
+    <Provider>
```
silva-fj: It seems we need to pass the theme here so the FAB uses our colors

ahoseinian: yes, thanks for noticing 👍

silva-fj (approved):

PR #626, opened 2021-11-24
Council voting

ishrak-imam:
### Description
Vote for council members.


src/screen/Council/CouncilScreen.tsx
```diff
@@ -132,17 +175,200 @@ function Item({
...
+  const [account, setAccount] = React.useState<string>();
+  const [amount, setAmount] = React.useState<string>('');
+  const [selectedCandidates, setSelectedCandidates] = React.useState<Array<string>>([]);
+
+  const {api} = useApi();
+  const startTx = useApiTx();
+  const {networkAccounts} = useAccounts();
+  const formatBalance = useFormatBalance();
+
+  const onCandidateSelect = (accountId: string, isSelected: boolean) => {
```
ahoseinian, line 201: probably something like this would work also, https://lodash.com/docs/4.17.15#xor
```suggestion
  const onCandidateSelect = (accountId: string, isSelected: boolean) => xor(selectedCandidates, [accountId])
```

src/screen/Council/CouncilScreen.tsx
```diff
@@ -132,17 +175,200 @@ function Item({
...
+
+  const disabled = !amount || !account || !(selectedCandidates.length > 0);
+
+  const bondValue = useMemo(() => {
+    if (api != null) {
+      const location = api.consts.elections || api.consts.phragmenElection || api.consts.electionsPhragmen;
+      return (
+        location &&
+        location.votingBondBase &&
+        location.votingBondBase.add(location.votingBondFactor.muln(selectedCandidates.length))
```
ahoseinian:
```suggestion
        location?.votingBondBase?.add(location.votingBondFactor.muln(selectedCandidates.length))
```

src/screen/Council/CouncilScreen.tsx
```diff
@@ -2,19 +2,33 @@ import {useWindowDimensions} from 'react-native';
 import Identicon from '@polkadot/reactnative-identicon';
 import {createMaterialTopTabNavigator} from '@react-navigation/material-top-tabs';
 import {useNavigation} from '@react-navigation/native';
-import {Divider, Text} from '@ui-kitten/components';
+import {Divider, Text, Card, Input, Modal} from '@ui-kitten/components';
```
ahoseinian: Maybe some of this items could come from paper?

src/screen/Council/CouncilScreen.tsx
ishrak-imam: This could work but here we have an extra logic. The selection should stop once `MAX_VOTES` is reached.

src/screen/Council/CouncilScreen.tsx
```diff
@@ -132,17 +173,197 @@ function Item({
...
+
+  const onCandidateSelect = (accountId: string, isSelected: boolean) => {
+    isSelected
+      ? setSelectedCandidates(selectedCandidates.filter((id) => id !== accountId))
+      : selectedCandidates.length < MAX_VOTES
+      ? setSelectedCandidates([...selectedCandidates, accountId])
+      : noop;
+  };
+
+  const disabled = !amount || !account || !(selectedCandidates.length > 0);
```
silva-fj:
```suggestion
  const disabled = !amount || !account || selectedCandidates.length === 0;
```

src/screen/Council/CouncilScreen.tsx
```diff
@@ -2,19 +2,33 @@ import {useWindowDimensions} from 'react-native';
...
+import globalStyles, {monofontFamily, standardPadding} from 'src/styles';
 import {MotionsScreen} from './MotionsScreen';
+import {ScrollView, TouchableOpacity} from 'react-native';
+import {useModuleElections} from 'src/api/hooks/useModuleElections';
+import Badge from 'presentational/Badge';
+import {noop} from 'lodash';
+import {useApiTx} from 'src/api/hooks/useApiTx';
+import {useTheme} from 'context/ThemeContext';
+
+const MAX_VOTES = 16;
```
silva-fj, line 31: I guess you got this value from the polkadotjs app, right? It seems like an arbitrary number. I wonder if we could get it from the api

src/screen/Council/CouncilScreen.tsx
```diff
@@ -132,17 +173,197 @@ function Item({
...
+          onFocus={() => setAmount('')}
+          onChangeText={(nextValue) => setAmount(nextValue.replace(/[^(\d+).(\d+)]/g, ''))}
+        />
+
+        <Text category="s1">{api ? formatBalance(getBalanceFromString(api, amount)) : ''}</Text>
+
+        <Padder scale={1} />
+
+        <View style={styles.centerAlign}>
+          <Text category="c1">Select upto 16 candidates in the preferred order:</Text>
```
silva-fj:
```suggestion
          <Text category="c1">{`Select up to ${MAX_VOTES} candidates in the preferred order:`}</Text>
```

src/screen/Council/CouncilScreen.tsx
silva-fj: What if the user has already voted with an account? Can we get from the chain if that's the case? Can he/she vote again?

src/screen/Council/CouncilScreen.tsx
ishrak-imam: It is hard coded in the polkadot app as well.

src/screen/Council/CouncilScreen.tsx
ishrak-imam: We can get for which council members user has already voted with an account and preselect them from the list but a user can vote multiple times for same council members using the same account.

src/screen/Council/CouncilScreen.tsx
```diff
@@ -30,67 +45,94 @@ export function CouncilScreen() {
...
+                }
+              }}
+            />
+          )}
+          ItemSeparatorComponent={Divider}
+          ListEmptyComponent={EmptyView}
+        />
+      )}
+      {council && moduleElection?.hasElections ? (
+        <CouncilVote
```
silva-fj: Let's call it `CouncilVoteModal` so it's a bit more clear

src/screen/Council/CouncilScreen.tsx
```diff
@@ -132,17 +174,205 @@ function Item({
...
+          onFocus={() => setAmount('')}
+          onChangeText={(nextValue) => setAmount(nextValue.replace(/[^(\d+).(\d+)]/g, ''))}
+        />
+
+        <Text category="s1">{api ? formatBalance(getBalanceFromString(api, amount)) : ''}</Text>
+
+        <Padder scale={1} />
+
+        <View style={styles.centerAlign}>
+          <Text category="c1">{`Select upto ${MAX_VOTES} candidates in the preferred order:`}</Text>
```
silva-fj:
```suggestion
          <Text category="c1">{`Select up to ${MAX_VOTES} candidates in the preferred order:`}</Text>
```
It should be **up to**

silva-fj (dismissed):

silva-fj (approved):

PR #683, opened 2021-12-14
Bug-669 Do not prompt PN permission automatically

### LINKED ISSUE #669: Bug : native permission for PN is shown before the permission prompt
Operating system : iOS

How to reproduce the bug:
- Reinstall the app on your iOS device 
- Launch the app
- Wait a bit for data to load

**What we currently see :**  A native dialog appears asking you to give permission to send PN and in the background the design created to be the permission prompt

**Expected behavior** : The prompt message is shown asking the user if he wants to give permissions to send him PN , if he says yes only then we show him the native dialog. See original ticket:  https://github.com/litentry/litentry-app/issues/163

ishrak-imam:
### Description

Closes https://github.com/litentry/litentry-app/issues/669

Previously in the `PermissionGrantingPrompt` component we were showing the native permission dialog for push notification automatically without user interaction as we did not store the permission skipped status. Which apparently led to this behaviour.

Now the skipped status is persisted in app storage, added skip functionality, and native dialog is prompted on user interaction only.

Additionally there are some refactoring like removing ui-kitten components from the `PermissionGrantingPrompt` screen


src/ui/navigation/AppNavigator.tsx
```diff
@@ -292,8 +291,10 @@ function AppNavigator() {
 
   return (
     <AppStack.Navigator screenOptions={overlayScreenOptions}>
-      {pushAuthorizationStatus && pushAuthorizationStatus === messaging.AuthorizationStatus.NOT_DETERMINED ? (
-        <AppStack.Screen name={routeKeys.permissionGrantingPromptScreen} component={PermissionGrantingPrompt} />
+      {isPnPromptNeeded ? (
+        <AppStack.Screen name={routeKeys.permissionGrantingPromptScreen}>
+          {() => <PermissionGrantingPrompt skipPnPermission={skipPnPermission} />}
```
silva-fj, line 296: Since this approach removes optimizations from react navigation to avoid re-renders you may want to use `React.memo` in `PermissionGrantingPrompt` or just call `usePushAuthorizationStatus` to get `skipPnPermission` instead of passing it as a prop

ishrak-imam: Here it is removing the optimisations because of the render callback, that's true. But if the `usePushAuthorizationStatus` hook is called again it does not behaves as expected. Because it is a different instance of the hook.

Technically this component would be rendered very few times in reality. So for now it should be ok with `React.memo` which I have just added. Ideally we would need a context here but for this use case I think it is good enough.

silva-fj (approved):

PR #686, opened 2021-12-15
Issue-670 Display account name instead of address

### LINKED ISSUE #670: Display account's name instead of account's address in "My Accounts"
> As a user I wanna easily find an account in my accounts list

Right now, even though the user has set up a name for an account, we are still displaying the address in the list of accounts. Let's change the address by the name set up by the user

ishrak-imam:
### Description

closes https://github.com/litentry/litentry-app/issues/670

The account name display logic is in the following order:
1. If identity information found show name from identity
2. If local account name found show `<account_name>(<account_address>)`
3. lastly, show shortened accountId(address) if nothing found

Additionally this PR includes some refactoring in the Accounts screen (mainly using `Menu` from RN paper instead of `OverflowMenu` from ui-kitten )


src/ui/screens/AccountsScreen.tsx
```diff
@@ -43,32 +43,67 @@ export function AccountsScreen({navigation}: Props) {
...
+                  onPress={() => {
+                    setSortMenuVisible(true);
+                  }}
+                  style={globalStyles.rowAlignCenter}>
+                  <Subheading>Sort by</Subheading>
+                  <Icon name="chevron-down" size={25} />
+                </TouchableOpacity>
+              }>
+              <Menu.Item
+                style={sortBy === 'name' ? {backgroundColor: colors.background} : undefined}
```
silva-fj: NIT: another way to handle this is to disable the Item `<Menu.Item disabled={sortBy === 'name'} ... />`, this will change the color and also will prevent to call `sortAccounts` again

src/ui/screens/AccountsScreen.tsx
```diff
@@ -201,31 +192,29 @@ const Buttons = ({navigation}: {navigation: Props['navigation']}) => {
...
+          onPress: () => navigation.navigate(importAccountScreen),
+        },
+        {
+          icon: 'plus',
+          label: 'Add External Account',
+          onPress: () => navigation.navigate(addAccountScreen),
+        },
+        {
+          icon: 'key-plus',
+          label: 'Generate New Seed',
```
silva-fj: I think we should change the wording here, most users don't know what a seed is. Probably "Create New Account" would be better.

ishrak-imam: May be also change `Import seed` to `Import account`?

silva-fj: Yeah that would be better as well 👍🏼

silva-fj (approved):
👍🏼

PR #691, opened 2021-12-16
Issue-657 New empty state

### LINKED ISSUE #657: Add Empty State to every "no data" screen
> As a user, when I see an empty screen I wanna know if it is a bug or there is simply no data to display

The following empty screen has to be implemented everywhere where we have no data to display (eg: no bounties, no motions...). Moreover, it has to replace the current empty state illustration we have on the **My Accounts** screen.
❗️: For screen with Call to Actions, keep the Call to Action underneath the illustration

**The illustration** :

ishrak-imam:
### Description

Closes https://github.com/litentry/litentry-app/issues/657


ahoseinian (approved):
LGTM

silva-fj (commented):
What about other lists that are often empty? like bounties and motions

ishrak-imam:
> What about other lists that are often empty? like bounties and motions

Those are using the `EmptyView` component already.

PR #697, opened 2021-12-20
Feature/508 keychain passwords

### LINKED ISSUE #508: Feature: Store the user's password in the keychain/keystore
When creating/importing accounts the user needs to define a password to encrypt the key pair. Then to be able to sign transactions the user needs to unlock the account (introduce the password). We should add the option to the user so they can store the password in the keychain/keystore so they can use biometrics (FaceID / TouchID) to unlock the accounts instead of having to type the password every time they need to make a transaction.

- Use the SecureKeychain service we already have.
- Store the password when creating/importing accounts.
- Retrieve the password from the keychain when signing transactions.

ahoseinian:
fixes #508 

I had to add different services (app_id + account_id) to store different passwords for different accounts.


src/service/SecureKeychain.ts
```diff
@@ -34,14 +38,14 @@ async function getSupportedBiometryType() {
...
   mmkvStorage.delete(SELECTED_ACCESS_CONTROL);
 
-  return Keychain.resetGenericPassword({service: defaultOptions.service});
+  return Keychain.resetGenericPassword({service});
 }
 
-async function getGenericPassword() {
-  const userCredentials = await Keychain.getGenericPassword();
+async function getGenericPassword(accountId: string) {
+  const userCredentials = await Keychain.getGenericPassword({service: getKeychainService(accountId)});
```
silva-fj: I think we should keep this service (SecureKeychain) generic, later we could use it as well to unlock the entire app for example. `getGenericPassword` could take a `service` string instead of the `accountId`

src/service/SecureKeychain.ts
```diff
@@ -22,6 +22,10 @@ const defaultOptions = {
   fingerprintPromptCancel: 'Cancel',
 };
 
+function getKeychainService(accountId: string) {
```
silva-fj: How about `serviceId` or just `service` to keep it generic?

src/service/SecureKeychain.ts
ahoseinian: I've updated the keychain file to export functions both for generic password & password by service id. please check it again.

silva-fj (approved):
cool 👍🏼

PR #726, opened 2021-12-22
Bug-701 Fix multiple back navigation if pressed quickly

### LINKED ISSUE #701: Bug: Stuck in the network connect screen
## Steps To Reproduce

1. From dashboard navigate to council screen
2. Try to go back to dashboard immediately by pressing the header back button multiple times

## The current behaviour
The app goes to dashboard

## The expected behaviour
The app goes to the `Connected to <network_name>` screen and the user is stuck there because of not having any way to interact with the app.

ishrak-imam:
### Description

Fixes https://github.com/litentry/litentry-app/issues/701

This PR includes the fix for the behaviour we see if back button is pressed multiple times in quick succession from the header. However, the other part of the problem is pressing android system back button which is addressed [here](https://github.com/litentry/litentry-app/pull/729)

### DIFF
--- a/src/ui/navigation/AppBars.tsx
+++ b/src/ui/navigation/AppBars.tsx
@@ -49,12 +49,19 @@ export function MainStackAppBar({
     }
   };
 
+  const goBack = () => {
+    const navIndex = navigation.getState().index;
+    if (navIndex > 0) {
+      navigation.goBack();
+    }
+  };
+
   return (
     <AppHeader style={{backgroundColor: colors.background}}>
       {options.headerLeft ? (
         options.headerLeft({onPress: openDrawer})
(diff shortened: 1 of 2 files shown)


ahoseinian (approved):
We wouldn't have the same problem with android back button being tapped twice?

silva-fj (approved):

ishrak-imam:
> We wouldn't have the same problem with android back button being tapped twice?

Yes, you are correct. Fixed it in a different [PR](https://github.com/litentry/litentry-app/pull/729)

ishrak-imam:
This PR is not needed anymore because of [this](https://github.com/litentry/litentry-app/pull/729)

PR #729, opened 2021-12-22
Issue-667 Remove API Loading Screen

### LINKED ISSUE #667: Report and Suggestion: Lazyload when connecting to the nodes
Currently, when we connect to the remote node, the "connecting" pages float on the top of the app. 

It happens mostly when 1. first loads the app.  2. switch back to the app. 

But when I do multi-tasks, it is common for me to switch back and forth among my apps. But every time when I switch back I need to wait a while, this is kind of annoying.

So it would be better to keep the existed information in the memory when I switched to some other apps, and update the current page when connected. Probably a "refresh" icon on the top would be enough to indicate we are reconnecting but not a new page to block all the information.

### ⚠️ what do we need?

- [x] UI/UX
- [x] Data Storage in Memory

ishrak-imam:
### Description

Fixes: https://github.com/litentry/litentry-app/issues/667

As discussed in the above thread we wanted to remove the API loading state and that would fix the android back press [issue](https://github.com/litentry/litentry-app/pull/726) automatically.

Additionally this PR includes loading state improvements in the Dashboard.


ahoseinian (approved):

src/context/ChainApiContext.tsx
```diff
@@ -114,18 +116,18 @@ type ChainApiContext = {
   reconnect: () => void;
 };
 
-export function useApiReconnect() {
-  const context = useContext(ChainApiContext);
+// export function useApiReconnect() {
```
silva-fj: In case you think we need to test something before cleaning this up, let's create an issue to clean this so we don't forget. You can add a todo with the link to the issue here

silva-fj (approved):

src/context/ChainApiContext.tsx
ishrak-imam: Cleaned it.

PR #756, opened 2021-12-29
Using github actions with fastlane to build and deploy to the stores

ahoseinian:


silva-fj:
@geastwood @ishrak-imam @chenzongxiong Can you guys take a quick look a this PR, I need to get it merged to test dependent workflows (they only run if the workflow file is in the default branch). Later we can work on improvements

geastwood (approved):
LGTM

PR #786, opened 2022-01-18
Remove unused components

ishrak-imam:
### Description

Some unused/redundant components were found in the code which are cleaned up with this PR.


src/ui/components/NetworkLogo.tsx
```diff
@@ -1,61 +0,0 @@
-import React from 'react';
-import {Image} from 'react-native';
-import PolkadotLog from 'image/Polkadot.png';
```
silva-fj, line 3: Let's create an issue to check what images we need and clean them up as well. The Polkadot's logo has changed so this is out dated

silva-fj (approved):

PR #829, opened 2022-01-30
Fix bounties list broken layout

ishrak-imam:
### Description

Before:

After:


silva-fj (approved):
For now we can have this to fix the UI glitch, but need a new design for the Cards in the lists

PR #832, opened 2022-02-01
Issue 827 mnemonic error feedback

### LINKED ISSUE #827: Bug: App doesn't show error message when mnemonic has been entered incorrectly
It seems like the app doesn’t give any response when there is an error in the mnemonic seed field (no error message)

ishrak-imam:
### Description

Fixes: https://github.com/litentry/litentry-app/issues/827

https://user-images.githubusercontent.com/16683923/151958235-c59f2555-0e41-4e30-a3ff-377b21a909a3.mp4

### DIFF
--- a/src/hooks/useKeyboard.ts
+++ b/src/hooks/useKeyboard.ts
@@ -1,0 +1,23 @@
+import React from 'react';
+import {Keyboard} from 'react-native';
+
+export function useKeyboard() {
+  const [isVisible, setIsVisible] = React.useState(false);
+
+  React.useEffect(() => {
+    const keyboardDidShow = Keyboard.addListener('keyboardDidShow', () => {
+      setIsVisible(true);
+    });
+
+    const keyboardDidHide = Keyboard.addListener('keyboardDidHide', () => {
+      setIsVisible(false);
+    });
+
+    return () => {
+      keyboardDidShow.remove();
+      keyboardDidHide.remove();
+    };
+  }, []);
+
+  return isVisible;
+}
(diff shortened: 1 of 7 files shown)


src/ui/screens/ImportAccountScreen.tsx
```diff
@@ -50,49 +49,36 @@ function ImportAccount({navigation}: {navigation: NavigationProp<AccountsStackPa
...
+      encoded,
+      meta: {
+        name: account.title,
+        network: currentNetwork.key,
+        isFavorite: false,
+      },
+      isExternal: false,
+    };
+    addAccount(newAcc);
+    SecureKeychain.setPasswordByServiceId(account.password, 'BIOMETRICS', _address);
```
ishrak-imam, line 80: @franjs 
I thought we were using the formatted address in the secure key chain for import account, no?

ishrak-imam, line 82: @franjs 
How about we use the formatted address for SecureKeyChain as well. I am testing the impact of this change in a device build. Will revert if it introduces bugs.

silva-fj: Yep, this should be fine 👍🏼

src/hooks/useKeyboard.ts
```diff
@@ -0,0 +1,23 @@
...
+      setIsVisible(false);
+    });
+
+    return () => {
+      keyboardDidShow.remove();
+      keyboardDidHide.remove();
+    };
+  }, []);
+
+  return isVisible;
```
silva-fj: To make it easier to understand what this hook does we could return `{ isVisible }`, or call the hook `useKeyboardStatus` and return `'visible' | 'hidden'`

silva-fj (approved):

Issue #857, opened 2022-02-10
Bug: the keyboard doesn't closed when clicked on Menu


Ekaanth:
## Steps To Reproduce

1. go to feedback screen> type a feedback
2. click on the many Icon/ header will not result in close of keyboard

Link or media file of the bug:

![Screenshot_1644502012](https://user-images.githubusercontent.com/22363352/153424933-a3b2c91e-c9d6-423c-b1bf-6f4e90b01f71.png)

## The expected behavior

The keyboard should dismiss on click of the menu or top navigation bar.

PR #858, opened 2022-02-10
852: Keyboard drop on textinput background click

Ekaanth:
Issue #857  

Screenshot_1644506457


ishrak-imam (approved):

silva-fj (commented):
This won't fix the issue if you open the drawer with the Keyboard opened.. You need to call `Keyboard.dismiss()` in the AppBar (when pressing the icon to open the drawer)

Ekaanth:
> 

Yes @franjs , even when I used `Keyboard.dismiss()` on TouchableWithoutFeedback component the keyboard is still shown. That's the reason I created a new issue for that. #857

silva-fj (approved):

PR #866, opened 2022-02-14
issue-837- fixed the funds to send in decimal

### LINKED ISSUE #837: Bug: Unable to send funds with decimals
As a user I would like to be able to send fractions of funds from my wallet. When trying to add `,` it gets deleted automatically.

File

## The current behavior
Currently only integer numbers can be added in the amount.

## The expected behavior
One should be able to send fractions of funds (e.g: 0.25)

Ekaanth:
Irrespective of the phone location, replace comma(,) to dot(.) also avoid multiple decimals

### DIFF
--- a/src/ui/screens/SendFundScreen.tsx
+++ b/src/ui/screens/SendFundScreen.tsx
@@ -14,7 +14,7 @@ import {Button, Headline, IconButton, Text, TextInput} from '@ui/library';
 import {Padder} from '@ui/components/Padder';
 import {Layout} from '@ui/components/Layout';
 import {standardPadding} from '@ui/styles';
-
+import {decimalKeypad} from 'src/utils';
 type Props = {
   navigation: NavigationProp<AccountsStackParamList, typeof sendFundScreen>;
   route: RouteProp<AccountsStackParamList, typeof sendFundScreen>;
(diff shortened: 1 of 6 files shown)


src/ui/screens/SendFundScreen.tsx
```diff
@@ -65,7 +66,7 @@ export function SendFundScreen({navigation, route}: Props) {
               keyboardType="decimal-pad"
               value={amount}
               onFocus={() => setAmount('')}
-              onChangeText={(nextValue) => setAmount(nextValue.replace(/[^(\d+).(\d+)]/g, ''))}
+              onChangeText={(nextValue) => setAmount(decimalKeypad(nextValue))}
```
silva-fj, line 69: I have tested this on my device and it does not work. Here we can just do `nextValue.replace(',', '.')`, the other regex is not needed since the keyboardType will not allow typing letters. We should also add `contextMenuHidden={true}` to disallow copy/paste any other string here

What we need to do in these forms is adding validation/error handling. If the amount or the address are not correct the transaction will throw an error

silva-fj (changes requested):
Since we are doing the same in multiple places, another approach instead of adding an utils function is to create component for this purpose like `BalanceInput` or `AmountInput`. But we can create a separate ticket for this.

src/ui/screens/SendFundScreen.tsx
Ekaanth: @franjs  for now, contextMenuHidden={true} has been added to avoid copy-paste, and the decimalKeypad method is improved

silva-fj (approved):

Issue #877, opened 2022-02-16
UX education - decimals 


ghost:

ghost:
Education around https://support.polkadot.network/support/solutions/articles/65000168651-what-is-the-existential-deposit-

ghost:
Example 
https://polkadot.subscan.io/account/16fDvYYTNCNkiA9MLiocPxvbd7S61rqAjTArJtuB4ohQUz4d?tab=transfer

PR #879, opened 2022-02-16
Feature/874 display maximum tokens present in the wallet

### LINKED ISSUE #874: Feature: Display the maximum amount of tokens present in the selected wallet.
While voting or transferring tokens, the user doesn't know the total tokens present in the wallet. In this feature, implement and display the total number of tokens.

Ekaanth:
Will now be able to see their maximum available tokens:

Simulator Screen Shot - iPhone 12 - 2022-02-15 at 17 48 47

Simulator Screen Shot - iPhone 12 - 2022-02-15 at 17 48 19

Simulator Screen Shot - iPhone 12 - 2022-02-15 at 17 48 25


ishrak-imam (approved):

src/ui/components/MaxBalance.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+function MaxBalance(props: PropTypes) {
+  return (
+    <View style={styles.balance}>
+      {!props.address ? <Caption>MAX: -- </Caption> : <AccountInfo address={props.address} />}
+    </View>
+  );
+}
+
+const AccountInfo = (props: any) => {
```
ishrak-imam: Let's use the exact type instead of `any`.

silva-fj (commented):
We should also validate that the amount entered is not higher than the address balance.
It can be a separate issue though

Ekaanth:
> We should also validate that the amount entered is not higher than the address balance. It can be a separate issue though

I think we should resolve that while validating the form, while the user has entered the correct address and right amount of balance, while we need to also discuss issue #877 to make it more concrete, what do you think about it?

silva-fj:
> It can be a separate issue

Yep, that's why I said that It can be added in a separate issue (ticket)

PR #884, opened 2022-02-17
Fetch Council overview from litentry graph

ishrak-imam:
### Description

- Fetch council overview from litentry graph
- Fix registrars query and style adjustments


src/ui/components/CouncilSummaryTeaser.tsx
```diff
@@ -5,17 +5,17 @@ import {Padder} from '@ui/components/Padder';
...
 import {AccountTeaser} from '@ui/components/Account/AccountTeaser';
+import {useCouncil} from 'src/api/hooks/useCouncil';
 
 type PropTypes = {
   onPress: () => void;
 };
 
 export function CouncilSummaryTeaser(props: PropTypes) {
-  const {data, loading} = useCouncilSummary();
+  const {data, loading} = useCouncil();
```
silva-fj: This will be good for caching the council data before navigating to the screen but If we use this query here the app will always fetch everything (members, runners up, etc) even if the user never goes to the council screen.

ishrak-imam: Yes, we can have a separate summary query which only fetches limited information.

src/api/hooks/useCouncilSummary.ts
```diff
@@ -0,0 +1,76 @@
...
+        termLeftParts
+        percentage
+      }
+    }
+  }
+`;
+
+const oneMinute = 60 * 1000;
+
+export function useCouncilSummary() {
```
silva-fj: We already have this hook, but for some reason it has `.tsx` extension. Can you change it to `.ts` and use that one? There are fields here that we don't need

silva-fj (approved):

PR #903, opened 2022-02-22
issue:898-Notifications after disabling: initial commit

Ekaanth:
Now the user will be able to Allow notification despite when opted skip in the initial process. 

Simulator Screen Shot - iPhone 12 - 2022-02-22 at 09 58 14

Simulator Screen Shot - iPhone 12 - 2022-02-22 at 09 58 20


silva-fj:
We should cover the following use cases to make it easier for the users:

- **User has skipped the initial permission prompt**:
When the user navigates to the notification settings screen, it should be prompted with the `PermissionGrantingPromptScreen`.

- **User has denied the permission:**
When the user navigates to the notification settings screen, it should be prompted with an alert saying that the notifications need to be enabled in the phone settings, When pressing OK it should open the notifications settings of the app (We can use Linking to open it), otherwise the Switches in the screen should be disabled (the user should not be able to toggle the switch) and a message should be displayed saying that PN needs to be enabled in the phone settings. A button to open the phone notifications settings should be displayed as well. The Alert should only be displayed once per session (if the user goes to another screen and navigates back to the, it should not trigger the alert, but if closes the app and it should be trigger again when navigating to the notifications settings).

Let me know if you have any questions.

PR #933, opened 2022-02-28
Fix add external account via QR Scanning

ishrak-imam:
### Description

Actually if we scan a public address the feature is working fine. However, scanning a QR containing an invalid address such as wrong format or the QR for the private key from parity signer will result into an exception which was unhandled and eventually causing the crash.

In this PR the exception was handled with an alert message.

@stephanie-olp Can we also add a message to the user that they need to scan the QR for the public address. The QR scanning feature to add external account is mainly there so that the user doesn't have to input the address manually. So it will have to be the public address always.

### DIFF
--- a/src/ui/screens/AddAccountScreen/AddAccountScreen.tsx
+++ b/src/ui/screens/AddAccountScreen/AddAccountScreen.tsx
@@ -64,6 +64,8 @@ export function AddAccountScreen({navigation}: {navigation: NavigationProp<AppSt
         if (isAddressValid(currentNetwork, parsed.address)) {
           dispatch({type: 'SET_ADDRESS', payload: parsed.address});
           dispatch({type: 'SET_STEP', payload: 'preview'});
+        } else {
+          Alert.alert('Validation Failed', 'Please scan the address from the correct network');
         }
       } catch (e) {
         Alert.alert('Validation Failed', 'Address is invalid.');
--- a/src/ui/screens/AddAccountScreen/AddAccountScreen.tsx
+++ b/src/ui/screens/AddAccountScreen/AddAccountScreen.tsx
@@ -1,4 +1,4 @@
-import React, {useCallback, useContext, useEffect, useMemo, useReducer, useRef} from 'react';
+import React, {useCallback, useContext, useEffect, useMemo, useReducer, useRef, useState} from 'react';
 import {Alert, StyleSheet, View} from 'react-native';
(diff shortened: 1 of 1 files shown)


src/ui/screens/AddAccountScreen/AddAccountScreen.tsx
```diff
@@ -112,7 +118,7 @@ export function AddAccountScreen({navigation}: {navigation: NavigationProp<AppSt
                       </TabScreen>
                       <TabScreen label="Via QR" icon="qrcode">
                         <View style={globalStyles.paddedContainer}>
-                          <QRCamera onRead={handleScan} />
+                          {tabIndex === 1 && <QRCamera onRead={handleScan} />}
```
ishrak-imam, line 123: This is to block the rendering of `QRCamera` component until navigated to that tab. Otherwise it scans any visible QR even if not navigated to the tab.

src/ui/screens/AddAccountScreen/AddAccountScreen.tsx
```diff
@@ -59,10 +59,14 @@ export function AddAccountScreen({navigation}: {navigation: NavigationProp<AppSt
 
   const handleScan = useCallback(
     ({data}) => {
-      const parsed = parseAddress(data);
-      if (isAddressValid(currentNetwork, parsed.address)) {
-        dispatch({type: 'SET_ADDRESS', payload: parsed.address});
-        dispatch({type: 'SET_STEP', payload: 'preview'});
+      try {
+        const parsed = parseAddress(data);
+        if (isAddressValid(currentNetwork, parsed.address)) {
```
silva-fj, line 64: When this is false, we should also display an error message as well, explaining that the address scanned is for another network

src/ui/screens/AddAccountScreen/AddAccountScreen.tsx
```diff
@@ -59,10 +59,16 @@ export function AddAccountScreen({navigation}: {navigation: NavigationProp<AppSt
...
+          dispatch({type: 'SET_ADDRESS', payload: parsed.address});
+          dispatch({type: 'SET_STEP', payload: 'preview'});
+        } else {
+          Alert.alert('Validation Failed', 'Please scan the address from the correct network');
```
silva-fj:
```suggestion
          Alert.alert('Validation Failed', `The address must belong to the ${currentNetwork.name} network.`);
```

silva-fj (approved):

PR #938, opened 2022-03-01
Fetching crowdloan data from the litentry-graph

### LINKED ISSUE #891: Crowdloan
Fetch Crowdloan data from the litentry-graph

silva-fj:
Fixes #891


src/ui/screens/Parachains/CrowdloanScreen.tsx
```diff
@@ -142,47 +112,39 @@ export function CrowdloanScreen() {
...
   const {colors} = useTheme();
+  const {currentNetwork} = useContext(NetworkContext);
 
-  const lastEndpoint = endpoints?.[endpoints.length - 1] as LinkOption;
-  const text = lastEndpoint?.text ?? `#${item.paraId.toString()}`;
-
-  const percentage = cap.isZero() ? 100 : raised.muln(10000).div(cap).toNumber() / 10000;
+  const isLitentryParachain = currentNetwork.key === 'polkadot' && item.paraId === '2013';
+  const isLitmusParachain = currentNetwork.key === 'kusama' && item.paraId === '2106';
+  const isSpecial = isLitentryParachain || isLitmusParachain;
```
ishrak-imam, line 122: This logic is already implemented in the backend. Can we add extra status flags from the graph?

ishrak-imam (approved):
LGTM!

src/ui/screens/Parachains/CrowdloanScreen.tsx
silva-fj: I was thinking that since this logic is only relevant for us it could leave in the app. We can change it later if we think other products can use it.

PR #951, opened 2022-03-07
Disabling layout animation on android (Events screen)

silva-fj:
There are some issues with layout animations in the current version of `react-native-reanimated` on android.


ishrak-imam (approved):
I was also thinking of displaying the calendar in weeks instead of months if possible.
It takes up almost all of the screens.

Can be a separate PR though.

silva-fj:
Yes we can definitely improve the screen, like have the option to export an event to the phone's calendar so people can have reminders, etc. We could also use the `Agenda` component to have the option to see the month in the cases there are not events in weeks. We need to do some planning and have proper UX design for this

PR #952, opened 2022-03-07
Fix address QR code scanner

### LINKED ISSUE #949: Bug: QR scanner doesn't work further after scanning wrong address
App version:

## Steps To Reproduce

1. Scan a wrong address or an address in the wrong network while trying to add external account via QR
2. Press ok on the error alert

## The current behaviour
The error alert is gone but the QR scanner doesn't scans further or provides any feedback

## The expected behaviour
The QR scanner gets ready for further scanning and gives feedback

** While fixing the add account part the QR scanner for registering sub-identities can also be fixed the same way.

silva-fj:
Fixes #949


ishrak-imam (approved):
LGTM!

PR #990, opened 2022-03-11
DisplayShort version

### LINKED ISSUE #948: Display a short version of balances when it makes sense
We are getting the formatted balances from the api, but in some places we want to display a shorter version:

IMG_1014

Ekaanth:
fix #948

### DIFF
--- a/src/ui/screens/Parachains/AuctionsScreen.tsx
+++ b/src/ui/screens/Parachains/AuctionsScreen.tsx
@@ -36,7 +36,7 @@ export function AuctionsScreen() {
 
   const {auctionsInfo, latestAuction} = auction;
   const {winningBid, leasePeriod, endingPeriod, raised, raisedPercent} = latestAuction;
-  const raisedFormatted = formatBalance(raised) ?? '';
+  const raisedFormatted = formatBalance(raised, {isShort: true}) ?? '';
   const remainingPercent =
     typeof endingPeriod?.remainingPercent === 'number' && endingPeriod.remainingPercent > 100
       ? 100
--- a/src/ui/screens/Parachains/AuctionsScreen.tsx
+++ b/src/ui/screens/Parachains/AuctionsScreen.tsx
@@ -12,11 +12,12 @@ import ProgressChartWidget from '@ui/components/ProgressWidget';
 import {EmptyView} from '@ui/components/EmptyView';
 import Clipboard from '@react-native-community/clipboard';
 import {useSnackbar} from 'context/SnackbarContext';
+import {useFormatBalance} from 'src/api/hooks/useFormatBalance';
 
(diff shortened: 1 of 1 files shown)


src/ui/screens/Parachains/AuctionsScreen.tsx
```diff
@@ -35,6 +36,7 @@ export function AuctionsScreen() {
 
   const {auctionsInfo, latestAuction} = auction;
   const {winningBid, leasePeriod, endingPeriod, raised, raisedPercent} = latestAuction;
+  const raisedFormatted = formatBalance(raised) ?? '';
```
ishrak-imam: You have to pass the `isShort` argument to get the short value.

```suggestion
  const raisedFormatted = formatBalance(raised, {isShort: true}) ?? '';
```

ishrak-imam (commented):
Can you also check if we need to display the short version anywhere else in the app.

Ekaanth:
> Can you also check if we need to display the short version anywhere else in the app.
I haven't come across any pages where the value overlaps on the graphs.

silva-fj (approved):

silva-fj:
> > Can you also check if we need to display the short version anywhere else in the app.
> > I haven't come across any pages where the value overlaps on the graphs.

You can just manually hard code a value to test it

Ekaanth:
> > > Can you also check if we need to display the short version anywhere else in the app.
> > > I haven't come across any pages where the value overlaps on the graphs.
> 
> You can just manually hard code a value to test it

In all of the applications, we display tokens in only one place, and that is here.

ishrak-imam (approved):

silva-fj:
> > > > Can you also check if we need to display the short version anywhere else in the app.
> > > > I haven't come across any pages where the value overlaps on the graphs.
> > 
> > 
> > You can just manually hard code a value to test it
> 
> In all of the applications, we display tokens in only one place, and that is here.

Not exactly, we display amounts in the crowdloan screen as well. In small screens in the crowdloan screen looks a bit odd when the amount is high, but I think we need to solve that issue in a different way there, it's important to display the total formatted amount

Ekaanth:
> > > > > Can you also check if we need to display the short version anywhere else in the app.
> > > > > I haven't come across any pages where the value overlaps on the graphs.
> > > 
> > > 
> > > You can just manually hard code a value to test it
> > 
> > 
> > In all of the applications, we display tokens in only one place, and that is here.
> 
> Not exactly, we display amounts in the crowdloan screen as well. In small screens in the crowdloan screen looks a bit odd when the amount is high, but I think we need to solve that issue in a different way there, it's important to display the total formatted amount

Despite showing the formatted value we still have the issue with displaying, it has more to do with the design and can be addressed when we tackle this issue.

PR #997, opened 2022-03-15
Setting up MSW to mock server data in tests/development

silva-fj:
This PR introduces [MSW](https://mswjs.io/) to mock server data. From now on we can start adding handlers that return mocked data (data we get from the litentry-graph). 

- Added mock data for registrars.
- Added a test example for the RegistrasList component.
- Added `testUtils` to expose the utility functions from react native testing library (wrapping the test components with providers).
- Added config to run tests on CI


src/testUtils.tsx
```diff
@@ -0,0 +1,32 @@
...
+import fetch from 'cross-fetch';
+
+// import {LitentryApiClientProvider} from 'context/LitentryApiContext';
+import ThemeProvider from 'context/ThemeContext';
+
+export const client = new ApolloClient({
+  cache: new InMemoryCache(),
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  link: new HttpLink({uri: 'http://localhost:3000/graphql', fetch}),
```
ishrak-imam, line 11: What does this local url mean? 
Is it like it doesn't have any impact since api hooks in test environment will get mocked data?

silva-fj: Correct, it does not matter what we put here, the client (ApolloClient) just need a valid uri to trigger requests and the requests will be intercepted by MSW

src/testUtils.tsx
```diff
@@ -0,0 +1,32 @@
...
+import {ApolloProvider, ApolloClient, InMemoryCache, HttpLink} from '@apollo/client';
+import fetch from 'cross-fetch';
+
+// import {LitentryApiClientProvider} from 'context/LitentryApiContext';
+import ThemeProvider from 'context/ThemeContext';
+
+export const client = new ApolloClient({
+  cache: new InMemoryCache(),
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
ishrak-imam: Why do we need the @ts-ignore here?

silva-fj: `HttpLink` expects a `fetch` with a different interface, we are using `cross-fetch` which has a simpler interface (less config options)

ishrak-imam: I don't see any TS errors in my machine 🤔 .
But it's fine.

ishrak-imam (approved):
Cool set-up. Long awaited test coverage is finally coming up. 🚀  🚀

Ekaanth (approved):

src/testUtils.tsx
silva-fj: Actually that was for `node-fetch`, another library recommended to do the same but I got some TS issues with the dependencies of that library on jest and changed it. We don't need to ignore anything here now

PR #1001, opened 2022-03-15
Add info text in tech committee screen

ishrak-imam:


Ekaanth:
We can add information to a new screen using an overlay screen. Then, we can add more information to that page in the future. IMO :))

ishrak-imam:
> We can add information to a new screen using an overlay screen. Then, we can add more information to that page in the future. IMO :))

If we have a long list of committee members that is actually good idea. But for now I guess this is ok. Let's get this through and have feedback from UX.

Ekaanth (approved):

PR #1004, opened 2022-03-16
Feature/877 keep account alive

### LINKED ISSUE #877: UX education - decimals

Ekaanth:
fix #877 

blocked by https://github.com/litentry/litentry-graph/pull/188

Simulator Screen Shot - iPhone 12 - 2022-03-16 at 14 31 19

Simulator Screen Shot - iPhone 12 - 2022-03-16 at 14 46 15

### DIFF
--- a/src/ui/components/BalanceInput.tsx
+++ b/src/ui/components/BalanceInput.tsx
@@ -10,11 +10,12 @@ import {formattedStringToBn} from 'src/api/utils/balance';
 type PropTypes = {
   account?: Account;
   onChangeBalance: (dispatch: string) => void;
+  initialBalance?: string;
 };
 
 export function BalanceInput(props: PropTypes) {
   const {account} = props;
-  const [amount, setAmount] = useState('');
+  const [amount, setAmount] = useState(props.initialBalance ?? '');
   const {formatBalance, stringToBn} = useFormatBalance();
 
   const hasEnoughBalance = useMemo(() => {
@@ -32,13 +33,15 @@ export function BalanceInput(props: PropTypes) {
       <TextInput
         dense
         style={styles.textInput}
-        error={!hasEnoughBalance}
+        error={Boolean(!hasEnoughBalance && amount)}
         mode="outlined"
         autoComplete="off"
         placeholder="Enter amount"
         keyboardType="decimal-pad"
         value={amount}
-        onFocus={() => setAmount('')}
+        onFocus={() => {
+          amount;
+        }}
         onChangeText={(nextValue: string) => {
           setAmount(decimalKeypad(nextValue));
           props.onChangeBalance(decimalKeypad(nextValue));
--- a/src/ui/components/BalanceInput.tsx
+++ b/src/ui/components/BalanceInput.tsx
@@ -39,9 +39,6 @@ export function BalanceInput(props: PropTypes) {
         placeholder="Enter amount"
         keyboardType="decimal-pad"
         value={amount}
-        onFocus={() => {
-          amount;
-        }}
         onChangeText={(nextValue: string) => {
           setAmount(decimalKeypad(nextValue));
           props.onChangeBalance(decimalKeypad(nextValue));
--- a/src/ui/screens/SendFundScreen.tsx
+++ b/src/ui/screens/SendFundScreen.tsx
@@ -50,12 +50,18 @@ export function SendFundScreen({navigation, route}: Props) {
 
   const isEnteredBalanceValid = useMemo(() => {
     const enteredBalance = stringToBn(amount) ?? BN_ZERO;
-    const isBalanceValid =
-      enteredBalance.gt(BN_ZERO) && enteredBalance.lt(formattedStringToBn(accountInfo?.balance.formattedFree));
-    return isBalanceValid;
-  }, [amount, accountInfo, stringToBn]);
+    if (isKeepAliveActive) {
+      const keepAliveBalance = formattedStringToBn(accountInfo?.balance.free).sub(
+        formattedStringToBn(chainInfo?.existentialDeposit),
+      );
+      return (
+        enteredBalance.gt(formattedStringToBn(chainInfo?.existentialDeposit)) && enteredBalance.lt(keepAliveBalance)
+      );
+    }
+    return enteredBalance.gt(BN_ZERO) && enteredBalance.lt(formattedStringToBn(accountInfo?.balance.free));
+  }, [amount, accountInfo, stringToBn, isKeepAliveActive, chainInfo]);
 
-  const isSendDisabled = !isEnteredBalanceValid || !toAddress || !isToAddressValid;
+  const isSendDisabled = !isEnteredBalanceValid || !isToAddressValid;
 
   return (
     <Modalize ref={ref} adjustToContentHeight onClose={navigation.goBack} closeOnOverlayTap>
@@ -125,7 +131,7 @@ export function SendFundScreen({navigation, route}: Props) {
                       params: [toAddress, _amountBN],
                     })
                       .then(() => {
-                        snackbar('Funds transfered');
+                        snackbar('Funds transferred');
                         navigation.goBack();
                       })
                       .catch(() => {
--- a/src/ui/screens/SendFundScreen.tsx
+++ b/src/ui/screens/SendFundScreen.tsx
@@ -40,7 +40,7 @@ export function SendFundScreen({navigation, route}: Props) {
 
   useEffect(() => {
     ref.current?.open();
-  }, [accountInfo]);
+  }, []);
 
   const isAccountValid = useMemo(() => {
     return to ? isAddressValid(currentNetwork, to) : false;
@@ -72,10 +72,7 @@ export function SendFundScreen({navigation, route}: Props) {
           <View style={styles.container}>
             <Headline>Send funds</Headline>
             <Padder scale={1} />
-            <InputLabel
-              label="Enter amount"
-              helperText="Type the amount you want to transfer. Note that you can select the unit on the right e.g sending 1 milli is equivalent to sending 0.001."
-            />
+            <InputLabel label="Enter amount" helperText="Type the amount you want to transfer" />
             <BalanceInput account={accountInfo} onChangeBalance={setAmount} />
             <Padder scale={1} />
             <InputLabel
--- a/src/ui/screens/SendFundScreen.tsx
+++ b/src/ui/screens/SendFundScreen.tsx
@@ -30,11 +30,11 @@ export function SendFundScreen({navigation, route}: Props) {
   const {data: accountInfo} = useAccount(address);
   const ref = useRef<Modalize>(null);
   const [amount, setAmount] = React.useState('');
-  const [to, setTo] = React.useState<string>();
+  const [toAddress, setToAddress] = React.useState<string>();
   const [scanning, setScanning] = React.useState(false);
   const startTx = useApiTx();
   const {data: chainInfo} = useChainInfo();
-  const [isKeepAlive, setisKeepAlive] = React.useState(true);
+  const [isKeepAliveActive, setIsKeepAliveActive] = React.useState(true);
   const {currentNetwork} = useContext(NetworkContext);
   const snackbar = useSnackbar();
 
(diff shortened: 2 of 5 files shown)


src/ui/screens/SendFundScreen.tsx
```diff
@@ -30,11 +34,20 @@ export function SendFundScreen({navigation, route}: Props) {
...
   const {data: chainInfo} = useChainInfo();
+  const [isKeepAlive, setisKeepAlive] = React.useState(true);
+  const {currentNetwork} = useContext(NetworkContext);
+  const snackbar = useSnackbar();
 
   useEffect(() => {
     ref.current?.open();
   }, []);
 
+  const isAccountValid = useMemo(() => {
```
silva-fj:
```suggestion
  const isToAddressValid = useMemo(() => {
```

src/ui/screens/SendFundScreen.tsx
```diff
@@ -30,11 +34,20 @@ export function SendFundScreen({navigation, route}: Props) {
...
+    return to ? isAddressValid(currentNetwork, to) : false;
```
silva-fj: Let's change this name `toAddress` instead of just `to`.

src/ui/screens/SendFundScreen.tsx
```diff
@@ -57,17 +70,40 @@ export function SendFundScreen({navigation, route}: Props) {
...
+              helperText="Scan a contact address or paste the address you want to send funds to."
+            />
             <TextInput
               autoComplete="off"
               placeholder="To"
               value={to}
               onChangeText={(nextValue) => setTo(nextValue)}
               right={<TextInput.Icon name="qrcode" onPress={() => setScanning(true)} />}
             />
+            {!isAccountValid && to ? <HelperText type={'error'}>Enter a valid address</HelperText> : null}
```
silva-fj:
```suggestion
            {!isToAddressValid ? <HelperText type={'error'}>Enter a valid address</HelperText> : null}
```

src/ui/screens/SendFundScreen.tsx
```diff
@@ -30,11 +34,20 @@ export function SendFundScreen({navigation, route}: Props) {
   const [scanning, setScanning] = React.useState(false);
   const startTx = useApiTx();
   const {data: chainInfo} = useChainInfo();
+  const [isKeepAlive, setisKeepAlive] = React.useState(true);
```
silva-fj: NIT: `isKeepAliveEnabled` or `isKeepAliveActive`

src/ui/screens/SendFundScreen.tsx
```diff
@@ -76,22 +112,25 @@ export function SendFundScreen({navigation, route}: Props) {
                     const _amountBN = stringToBn(chainInfo.registry, amount);
                     startTx({
                       address,
-                      txMethod: 'balances.transferKeepAlive',
+                      txMethod: `${isKeepAlive ? `balances.transferKeepAlive` : `balances.transferKeepAlive`}`,
```
silva-fj:
```suggestion
                      txMethod: `${isKeepAlive ? `balances.transferKeepAlive` : `balances.transfer`}`,
```

src/ui/screens/SendFundScreen.tsx
```diff
@@ -76,22 +112,25 @@ export function SendFundScreen({navigation, route}: Props) {
...
                       params: [to, _amountBN],
                     })
                       .then(() => {
+                        snackbar('Funds transfered');
                         navigation.goBack();
                       })
                       .catch((e: Error) => {
                         console.warn(e);
```
silva-fj: If this fails we should inform the user that something went wrong.

src/ui/screens/SendFundScreen.tsx
```diff
@@ -57,17 +70,40 @@ export function SendFundScreen({navigation, route}: Props) {
...
             <TextInput
               autoComplete="off"
               placeholder="To"
-              value={to}
-              onChangeText={(nextValue) => setTo(nextValue)}
+              value={toAddress}
+              onChangeText={(nextValue) => setToAddress(nextValue)}
               right={<TextInput.Icon name="qrcode" onPress={() => setScanning(true)} />}
             />
+            {!isToAddressValid ? <HelperText type={'error'}>Enter a valid address</HelperText> : null}
```
ishrak-imam: NIT 💅 
```suggestion
            {!isToAddressValid ? <HelperText type="error">Enter a valid address</HelperText> : null}
```

And for other cases also.

src/ui/screens/SendFundScreen.tsx
```diff
@@ -133,6 +172,15 @@ const styles = StyleSheet.create({
     alignItems: 'center',
     borderRadius: 10,
   },
+  keepAlive: {
+    flexDirection: 'row',
+    justifyContent: 'flex-end',
+    paddingVertical: 15,
```
ishrak-imam: NIT 💅 

We could use the standard padding or a multiple of that.

```suggestion
    paddingVertical: standardPadding * 2,
```

src/ui/screens/SendFundScreen.tsx
```diff
@@ -133,6 +172,15 @@ const styles = StyleSheet.create({
...
+  },
+  keepAliveContainer: {
+    paddingVertical: 5,
+    paddingHorizontal: 10,
```
ishrak-imam:
```suggestion
    paddingHorizontal: standardPadding,
```

ishrak-imam (approved):
LGTM !!!

Awesome 🎉

silva-fj (approved):

Ekaanth:
@franjs @ishrak-imam  I have an additional requirement here, I believe, to validate the form when the user is set the toggler to `keepAlive` the entered value should be one less than the `formattedfree` correct me if I'm wrong here.

ishrak-imam:
> @franjs @ishrak-imam I have an additional requirement here, I believe, to validate the form when the user is set the toggler to `keepAlive` the entered value should be one less than the `formattedfree` correct me if I'm wrong here.

Yes, you are correct.

Ekaanth:
> > @franjs @ishrak-imam I have an additional requirement here, I believe, to validate the form when the user is set the toggler to `keepAlive` the entered value should be one less than the `formattedfree` correct me if I'm wrong here.
> 
> Yes, you are correct.

@ishrak-imam  Thanks for confirming. In that case when the stay alive is active what is the maximum token that can be transferred. how do I handle it?

ishrak-imam:
> > > @franjs @ishrak-imam I have an additional requirement here, I believe, to validate the form when the user is set the toggler to `keepAlive` the entered value should be one less than the `formattedfree` correct me if I'm wrong here.
> > 
> > 
> > Yes, you are correct.
> 
> @ishrak-imam Thanks for confirming. In that case when the stay alive is active what is the maximum token that can be transferred. how do I handle it?

We can pass the keepAlive configuration to the `<BalanceInput />` component and do the checking there. Also the `<BalanceInput />` component should actually return something like a `isValid` status along with the input amount.

If we think from one step back, The `<BalanceInput />` component can encapsulate the keepAlive toggle and all the logic in itself. It can only sent out the input amount and the `hasEnoughFunds` status to the parent.

Ekaanth:
> > > > @franjs @ishrak-imam I have an additional requirement here, I believe, to validate the form when the user is set the toggler to `keepAlive` the entered value should be one less than the `formattedfree` correct me if I'm wrong here.
> > > 
> > > 
> > > Yes, you are correct.
> > 
> > 
> > @ishrak-imam Thanks for confirming. In that case when the stay alive is active what is the maximum token that can be transferred. how do I handle it?
> 
> We can pass the keepAlive configuration to the `<BalanceInput />` component and do the checking there. Also the `<BalanceInput />` component should actually return something like a `isValid` status along with the input amount.
> 
> If we think from one step back, The `<BalanceInput />` component can encapsulate the keepAlive toggle and all the logic in itself. It can only sent out the input amount and the `hasEnoughFunds` status to the parent.

In this case, we can send the bond value to the component as well. My question here is how do we determine the token's value? For example an account has 10DOT and they want to transfer keeping their account alive, it should be formatedFree-bondValue-fee right? how do we do this?

silva-fj:
> > > > > @franjs @ishrak-imam I have an additional requirement here, I believe, to validate the form when the user is set the toggler to `keepAlive` the entered value should be one less than the `formattedfree` correct me if I'm wrong here.
> > > > 
> > > > 
> > > > Yes, you are correct.
> > > 
> > > 
> > > @ishrak-imam Thanks for confirming. In that case when the stay alive is active what is the maximum token that can be transferred. how do I handle it?
> > 
> > 
> > We can pass the keepAlive configuration to the `<BalanceInput />` component and do the checking there. Also the `<BalanceInput />` component should actually return something like a `isValid` status along with the input amount.
> > If we think from one step back, The `<BalanceInput />` component can encapsulate the keepAlive toggle and all the logic in itself. It can only sent out the input amount and the `hasEnoughFunds` status to the parent.
> 
> In this case, we can send the bond value to the component as well. My question here is how do we determine the token's value? For example an account has 10DOT and they want to transfer keeping their account alive, it should be formatedFree-bondValue-fee right? how do we do this?

You can take a look at the logic in the polkadotjs app [here](https://github.com/polkadot-js/apps/blob/master/packages/page-accounts/src/modals/Transfer.tsx)

ishrak-imam:
> > > > > @franjs @ishrak-imam I have an additional requirement here, I believe, to validate the form when the user is set the toggler to `keepAlive` the entered value should be one less than the `formattedfree` correct me if I'm wrong here.
> > > > 
> > > > 
> > > > Yes, you are correct.
> > > 
> > > 
> > > @ishrak-imam Thanks for confirming. In that case when the stay alive is active what is the maximum token that can be transferred. how do I handle it?
> > 
> > 
> > We can pass the keepAlive configuration to the `<BalanceInput />` component and do the checking there. Also the `<BalanceInput />` component should actually return something like a `isValid` status along with the input amount.
> > If we think from one step back, The `<BalanceInput />` component can encapsulate the keepAlive toggle and all the logic in itself. It can only sent out the input amount and the `hasEnoughFunds` status to the parent.
> 
> In this case, we can send the bond value to the component as well. My question here is how do we determine the token's value? For example an account has 10DOT and they want to transfer keeping their account alive, it should be formatedFree-bondValue-fee right? how do we do this?

The only responsibility of the `BalanceInput` component is to check if the entered amount is less than or equal to the free balance of the selected account. And if keep alive is enabled entered amount should be less than or equal to freeBalance - existential deposit.

For the voting bond according to polkadot js app the keepAlive check is not needed. So we control if we need the keepAlive feature in BalanceInput or not.

We can introduce an UI hook here with following pattern:

`const {amount, hasEnoughFunds, BalanceInputComponent} = useBalanceInput(isKeepAliveNeeded: true/false)`

ishrak-imam:
We will also need the unformatted existential deposit from the API to do the comparison.

src/ui/components/BalanceInput.tsx
```diff
@@ -32,13 +33,15 @@ export function BalanceInput(props: PropTypes) {
...
-        error={!hasEnoughBalance}
+        error={Boolean(!hasEnoughBalance && amount)}
         mode="outlined"
         autoComplete="off"
         placeholder="Enter amount"
         keyboardType="decimal-pad"
         value={amount}
-        onFocus={() => setAmount('')}
+        onFocus={() => {
+          amount;
```
silva-fj: What's this?

src/ui/screens/SendFundScreen.tsx
```diff
@@ -26,15 +32,31 @@ export function SendFundScreen({navigation, route}: Props) {
...
+  }, [toAddress, currentNetwork]);
+
+  const isEnteredBalanceValid = useMemo(() => {
+    const enteredBalance = stringToBn(amount) ?? BN_ZERO;
+    const isBalanceValid =
+      enteredBalance.gt(BN_ZERO) && enteredBalance.lt(formattedStringToBn(accountInfo?.balance.formattedFree));
+    return isBalanceValid;
+  }, [amount, accountInfo, stringToBn]);
+
+  const isSendDisabled = !isEnteredBalanceValid || !toAddress || !isToAddressValid;
```
silva-fj: You can remove `!toAddress` here, `isToAddressValid` will be false if `toAddress` is `undefined`

PR #1005, opened 2022-03-17
Feature/639 calendar teaser

### LINKED ISSUE #639: Feature: Calendar Teaser ("Upcoming events" section on the dashboard)
> As a user, when I open the app I quickly wanna find out what are the next events in Polkadot/ Kusama governance ecosystem

On the very top of the screen, let's add a section called "Upcoming events":
We can display the next 3 upcoming events on that list
If user clicks on that section he will be redirected to the calendar screen

* Wireframe: 
image

Ekaanth:
Fix #639


src/ui/screens/EventsCalendarTeaser.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export function EventsCalendarTeaser(props: Props) {
+  const {colors} = useTheme();
+  const {data: events, loading} = useCalendarEvents();
+  const topThreeEvents = events?.slice(0, 3);
+  const formateDate = (date: string) => {
+    const formatedDate = new Date(date);
+    const [eventDate, eventMonth] = [formatedDate.getDate(), formatedDate.getMonth()];
+    return eventDate.toString().concat('/').concat(eventMonth.toString());
+  };
```
ishrak-imam: Independent util methods like this can be outside the component so that these are not affected by component re-renders.

But in this case we should actually make use of the `dayjs` util and the method should live in `utils/date.ts`

ishrak-imam:
@Ekaanth Can you please add a screenshot with the empty state also?

Ekaanth:
Screenshot with the empty list.
UPDATED

src/utils/date.ts
```diff
@@ -10,3 +10,15 @@ export function fromNow(date: string) {
...
+  const currentDate = dayjs(date);
+  if (!format) {
+    return [currentDate.date(), currentDate.month().toString(), currentDate.year().toString()].join('/');
+  }
+  const formatValues = format.toUpperCase().split('/');
+  const mappedDates = formatValues.map((event) =>
+    event === 'DD' ? currentDate.date() : event === 'MM' ? currentDate.month() : currentDate.year(),
+  );
+  return mappedDates.join('/');
+}
```
ishrak-imam: Is there any specific reason we need to format it like this?

If I understood correctly, the API just returns dates as string and we can apply the formatting like following.

```suggestion
export function format(date: string, format = "DD/MM/YYYY") {
  return dayjs(date).format(format);
}
```

And call the method like `format("2022-03-21T15:36:19.410Z", "DD/MM")`

Ekaanth: Nope, I did not see how it worked. However, I'll keep this in mind.

src/ui/screens/AddBountyScreen.tsx
```diff
@@ -39,22 +39,22 @@ export function AddBountyScreen({navigation}: {navigation: NavigationProp<AppSta
 
   const {calculatedBountyBond, bountyAllocationBN, isBountyValid, isBountyTitleValid} = useMemo(() => {
     if (bounty && bounty.bountyDepositBase) {
-      const calculatedBountyBond = calculateBountyBond(
+      const bountyBondCalculated = calculateBountyBond(
```
silva-fj, line 42: Are the changes in this file part of this PR?

src/ui/screens/EventsCalendarTeaser.tsx
```diff
@@ -0,0 +1,75 @@
...
+import {Text, useTheme, Divider, Headline, Title} from '@ui/library';
+import {standardPadding} from '@ui/styles';
+import {getFormatedDate} from 'src/utils/date';
+import EmptyList from 'image/EmptyList.png';
+
+type Props = {
+  onPress: () => void;
+};
+
+export function EventsCalendarTeaser(props: Props) {
```
silva-fj: This is not a Screen, it should live in `@ui/components`

src/ui/screens/EventsCalendarTeaser.tsx
```diff
@@ -0,0 +1,75 @@
...
+          </React.Fragment>
+        ))
+      ) : (
+        <View style={styles.container}>
+          <View style={styles.imageContainer}>
+            <Image source={EmptyList} resizeMode="contain" />
+          </View>
+          <View style={styles.emptyContainer}>
+            <Headline>No Upcoming Events</Headline>
+            <Title style={{color: colors.primary}}>Check back soon</Title>
```
silva-fj: Making this primary will make it look like a link, also I don't think we need this message

src/ui/screens/EventsCalendarTeaser.tsx
```diff
@@ -0,0 +1,75 @@
...
+            <Divider />
+          </React.Fragment>
+        ))
+      ) : (
+        <View style={styles.container}>
+          <View style={styles.imageContainer}>
+            <Image source={EmptyList} resizeMode="contain" />
+          </View>
+          <View style={styles.emptyContainer}>
+            <Headline>No Upcoming Events</Headline>
```
silva-fj: `Headline` is probably too big to use it here, it's bigger than the section's title.. Maybe try `Subheading` or just `Text`

src/ui/screens/AddBountyScreen.tsx
Ekaanth: No, but I was getting lint warnings, which were supposed to be fixed in adding bounty screen.

silva-fj (approved):

PR #1083, opened 2022-04-07
Issue 1057 validation of forms in the app

### LINKED ISSUE #1057: Validation of forms in the app
- [ ] Go through all the forms and identify where it's needed
- [ ] Add validation to these places

### LINKED ISSUE #1069: Bug: Accounts screen unable to scroll
On click of any account details, unable to scroll.

## Steps To Reproduce

1. click on any user to redirect to the account screen
2. unable to scroll to balance card and account roles.

Ekaanth:
fixed #1069 and #1057 

Council screen:

crowdloan screen:

Referendum model


src/ui/screens/Council/CouncilScreen.tsx
```diff
@@ -229,7 +231,12 @@ function CouncilVoteModal({visible, setVisible, candidates, moduleElection}: Cou
       : noop;
   };
 
-  const disabled = !amount || !account || selectedCandidates.length === 0;
+  const isValidEnteredBalance = useMemo(() => {
```
ishrak-imam: Nit 💅 
```suggestion
  const isEnteredBalanceValid = useMemo(() => {
```

ishrak-imam (approved):
LGTM

Ekaanth:
Update :

src/ui/components/SubmitProposal.tsx
```diff
@@ -29,7 +32,13 @@ export function SubmitProposal() {
     dispatch({type: 'RESET'});
   };
 
-  const isDisabled = !state.account || !state.preimageHash || !state.balance;
+  const enteredBalance = stringToBn(state.balance ?? '0') ?? BN_ZERO;
```
silva-fj, line 35: Why do you need to add `?? BN_ZERO` ? `stringToBn` returns `BN`

src/ui/components/SubmitProposal.tsx
```diff
@@ -29,7 +32,13 @@ export function SubmitProposal() {
...
+  const democracyMinimumDeposit = formattedStringToBn(chainInfo?.democracyMinimumDeposit ?? '0') ?? BN_ZERO;
```
silva-fj: Same here

silva-fj (approved):
Good work 👍🏼

src/ui/components/SubmitProposal.tsx
Ekaanth: in this case `state.balance` return string | undefined despite it being set to string. since I'm not comparing the value of undefined so I used `??` to handle the undefined cases.

silva-fj: I understand that `state.balance` can be `undefined`, but the result of` stringToBn(state.balance ?? '0') ` would not be `undefined`, so `?? BN_ZERO` should not be needed.

Ekaanth: The result from `stringToBn` returns `BN` or `undefined`, thereby I had to do it like that. in another case, `formattedStringToBn` returns only `BN` I fixed for that one.

silva-fj: How can `stringToBn` return `undefined`? Here is the definition: 

`export function stringToBn(registry: Registry, input: string): BN `

Ekaanth: In this case, I'm using `stringToBn` from the `useFormatBalance` hook and not from Utils, but under the hood, it uses the `StringToBn` from utils.

silva-fj: Got it

silva-fj (approved):

PR #1084, opened 2022-04-07
Optimize council query

ishrak-imam:
### Description

In order to optimize the council query we had to request limited information from the chain for an account. The idea is to request for limited information for an account when it is meant to be displayed in a list and fetch the full account information in the details screen.

While doing that we had to make some values in the account object optional. So this PR includes the following changes:

- Update council summary and council queries/UI accordingly because of the schema change.
- Modify the AccountTeaser component to make is generic, so that it can be used as a list item too.
- Adjust the logic since some properties in the account object are optional now.
- Refactor the screen navigations to account details screen (always pass the callback from parent).


src/ui/components/Account/AccountTeaser.tsx
```diff
@@ -1,35 +1,43 @@
...
   identiconSize?: number;
+  onPress?: () => void;
+  children?: React.ReactNode;
+  testID?: string;
 };
 
-export function AccountTeaser({account, identiconSize = 20}: Props) {
-  const navigation = useNavigation<NavigationProp<AppStackParamList>>();
+export function AccountTeaser({account, onPress, children, testID, identiconSize = 20}: Props) {
+  if (onPress) {
```
silva-fj: Instead of this you can do `<TouchableOpacity disabled={!onPress} ... >`

src/ui/screens/Council/CouncilScreen.tsx
```diff
@@ -50,69 +52,74 @@ export function CouncilScreen() {
...
-              <List.Item style={styles.sectionHeader} title={buildSectionHeaderTitle(title, council)} />
-              {title === 'Members' ? (
-                <View style={globalStyles.rowContainer}>
+              <Padder />
+              <Subheading>{buildSectionHeaderTitle(title, council)}</Subheading>
+              {title === 'Member' ? (
+                <View style={styles.voteActions}>
                   <Button icon="vote" mode="outlined" onPress={() => setCouncilVoteVisible(true)}>
-                    Vote
+                    {`Vote`}
```
silva-fj, line 115: Why adding this?

src/ui/screens/Council/CouncilScreen.tsx
```diff
@@ -50,69 +52,74 @@ export function CouncilScreen() {
...
                   </Button>
                   <Padder scale={1} />
                   <Button icon="vote" mode="outlined" onPress={() => setSubmitCandidacyVisible(true)}>
-                    Submit Candidacy
+                    {`Submit candidacy`}
```
silva-fj, line 119: same

src/ui/screens/Council/CouncilScreen.tsx
ishrak-imam: Just personal preference. Also don't have any strong opinion.

silva-fj (approved):

PR #1086, opened 2022-04-08
Feature/1078 input address

### LINKED ISSUE #1078: Feature: Input address (scan and paste)
Wherever we have to enter a new address

Ekaanth:
FIX: #1078

Unable to test the scan on both IOS and Andriod

https://user-images.githubusercontent.com/22363352/162452198-4c422ef2-c72e-4cd6-b50e-c05b44f8e853.mp4


silva-fj:
Good work @Ekaanth, I have refactored a couple of things. I still think the UI for the QR code scanner is not ideal, we can iterate on this later with Miki.

This is how it looks now:
File (1)

silva-fj (approved):

Ekaanth:
> Good work @Ekaanth, I have refactored a couple of things. I still think the UI for the QR code scanner is not ideal, we can iterate on this later with Miki.
> 
> This is how it looks now: File (1)
Like @mikuczi  mentioned earlier on slack we need to provide a close icon button on the top right instead of a close text button, with the current UI it feels like an overlay screen. Providing a close button on the top right won't make much sense.

silva-fj:
> Like @mikuczi mentioned earlier on slack we need to provide a close icon button on the top right instead of a close text button, with the current UI it feels like an overlay screen. Providing a close button on the top right won't make much sense.

We need to change the pattern that we are using in the send funds screen altogether, but we can iterate on it with another ticket.

PR #1088, opened 2022-04-11
Upgrading react native paper to v4.12.0

silva-fj:
The issue with the `Card` component when changing the background color has been fixed in the new version. We no longer need the workaround.


ishrak-imam (approved):
Cool 👍

PR #1089, opened 2022-04-11
Replacing bn.js with a native implementation (react-native-bignumber)

silva-fj:
This library replaces `bn.js` with a native implementation in c++ using JSI. 

It would be a good idea add some performance monitoring into our app to be able to measure properly this kind of improvements. I'll tackle this in a separate PR

### DIFF
--- a/ios/Litentry.xcodeproj/project.pbxproj
+++ b/ios/Litentry.xcodeproj/project.pbxproj
@@ -249,6 +249,7 @@
 				00E356EB1AD99517003FC87E /* Frameworks */,
 				00E356EC1AD99517003FC87E /* Resources */,
 				C16081AA9903B5C63812BAAC /* [CP] Copy Pods Resources */,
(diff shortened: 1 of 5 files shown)


geastwood:
Should we go ahead with this?

Ekaanth (approved):

Ekaanth:
> Should we go ahead with this?

Looks good to me just curious if the package is entirely ready for production.

ishrak-imam:
> Should we go ahead with this?

I remember @franjs was willing to wait to check some metrics to see if it is really improving any performance.

After merging this https://github.com/litentry/litentry-app/pull/1092 there should be some metrics already and when this is merged we can compare. IIRC this was the reason to wait.

silva-fj:
Yeah the plan is to compare the performance metrics to see if we get any benefit from this because it was not noticeable. I'll take a look at this soon, we can measure this locally, there is no need to wait until this is merged to see the metrics on firebase

silva-fj:
closing in favor of https://github.com/litentry/litentry-app/pull/1133

PR #1092, opened 2022-04-11
Performance monitoring

silva-fj:
Installing and setting up [firebase performance monitoring](https://firebase.google.com/products/performance) and adding `react-native-startup-trace` to measure startup of the app.

As part of this PR I have also fixed the support for Flipper in our app.

@ishrak-imam @Ekaanth make sure to do the following after this is merged:
- Run .`/gradlew clean` in the android directory.
- Run `pod install --repo-update` in the ios directory.

I have also installed https://github.com/bamlab/react-native-performance so we can measure performance with Flipper


ishrak-imam (approved):
Looks good.

Do we need to install any flipper plugin?

silva-fj:
> Looks good.
> 
> Do we need to install any flipper plugin?

Yep, the one in the description (`rn-perf-monitor`)

Ekaanth (approved):
Looks good 👍

PR #1097, opened 2022-04-18
Balance input test case

Ekaanth:
Test case for balances component


src/ui/components/Balances.test.tsx
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+import React from 'react';
+import Balances from '@ui/components/Balances';
+import {create} from 'react-test-renderer';
+import {render, within} from 'src/testUtils';
+
+jest.doMock('src/ui/components/Balances', () => {
```
silva-fj: You should not mock the function/component under test.

src/ui/components/Balances.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return {
+    BalancesMockComponent: jest.fn(() => <Balances balance={accountBalanceDetails} />),
+  };
+});
+
+const {BalancesMockComponent} = require('src/ui/components/Balances');
+
+describe('Balances component', () => {
+  it('should render the prop component correctly', async () => {
+    const {getByTestId} = render(<BalancesMockComponent />);
```
silva-fj: As I mentioned last week, query by `testID` should be avoided if possible, you should try to query for the things that the user will see

silva-fj (commented):
Don't spend much time in this kind of tests, as I said they don't provide much value

Ekaanth:
I am closing this PR since this PR doesn't add any benefit to the app test cases.

PR #1099, opened 2022-04-18
issue-1074-remove unused dependencies

### LINKED ISSUE #1074: Remove unused dependencies
There are a few dependencies that we are not using anymore. Let's remove them

Ekaanth:
Fix for #1074 

By using [depcheck](https://www.npmjs.com/package/depcheck) package I was able to find out the unused packages present in the project, including the dev dependencies. With the help of depcheck I was able to remove all the unused dependencies but when I run `yarn` it reinstalls all the dependencies. we also have a few vulnerable from dependencies we can look into that high severity vulnerabilities

@ishrak-imam @franjs can you help me understand why the removed packages are getting installed back on `yarn`.


ishrak-imam:
@Ekaanth 
Not sure what you mean by the deps get back installed again. I tried running the branch locally and found that we actually need the `react-native-permissions` package for camera permissions. Apart from that everything seems fine.

Please check on real devices if possible or let me know if you need any help.

Ekaanth:
@ishrak-imam @franjs, Some packages have downgraded versions, I Initially checked with `depcheck` and `yarn remove` all the unused dependencies, after which I use `yarn install` and installed the packages again. Thereby getting the packages downgraded.

package.json
```diff
@@ -29,15 +29,11 @@
...
     "@polkadot/api": "7.9.1",
-    "@polkadot/apps-config": "^0.103.1",
     "@polkadot/reactnative-identicon": "0.89.2",
-    "@polkadot/ui-settings": "^0.89.2",
     "@react-native-community/async-storage": "^1.12.1",
     "@react-native-community/clipboard": "^1.5.1",
-    "@react-native-community/masked-view": "^0.1.11",
     "@react-native-firebase/app": "^14.7.0",
     "@react-native-firebase/messaging": "^14.7.0",
-    "@react-native-firebase/perf": "^14.7.0",
```
ishrak-imam, line 40: This should not be removed. I think it is needed for the new performance monitoring set-up we have.

ishrak-imam (commented):
About the other comment regarding dependencies getting downgraded, those deps are being installed because of the `nodeify` that we have in our scripts.

`nodeify` installs deps in specific versions and later in time dependabot updates those from time to time. Now if we remove those `nodeify` will install them again but will be a downgraded version.

I don't think an automatic check for unused dependencies will suit our project at the moment.

Also the deps that get reinstalled automatically should not be removed in the first place.

package.json
```diff
@@ -53,15 +50,14 @@
     "console-browserify": "^1.2.0",
     "dayjs": "^1.10.8",
     "ethers": "^5.6.1",
-    "events": "^3.3.0",
+    "events": "^1.1.1",
```
ishrak-imam: The dependencies that are automatically installed by `rn-nodify` are getting downgraded in the process. Those should not be removed in the first place.

package.json
```diff
@@ -91,7 +85,7 @@
     "react-native-permissions": "^3.3.1",
     "react-native-popable": "^0.4.3",
     "react-native-qrcode-scanner": "^1.5.5",
-    "react-native-randombytes": "^3.6.0",
+    "react-native-randombytes": "^3.6.1",
```
ishrak-imam, line 88: Same here.

package.json
```diff
@@ -106,10 +100,10 @@
     "react-native-vector-icons": "^9.1.0",
     "react-native-webview": "^11.17.2",
     "react-query": "^3.34.19",
-    "readable-stream": "^3.6.0",
+    "readable-stream": "^1.0.33",
```
ishrak-imam: Here.

package.json
```diff
@@ -106,10 +100,10 @@
...
     "rn-nodeify": "^10.3.0",
-    "stream-browserify": "^3.0.0",
-    "url": "^0.11.0",
+    "stream-browserify": "^1.0.0",
+    "url": "^0.10.3",
```
ishrak-imam: And here.

ishrak-imam (approved):
Other than that looks good to me. Please test test by running the app and checking the features.

package.json
Ekaanth: Yes!! I reverted the change, when using the `rn-nodify` it is automatically downgrading, is there something that needs to be looked at? or should I keep the same downgraded versions?

silva-fj: > should I keep the same downgraded versions?

No, please keep the versions as they currently are

silva-fj (commented):
You also need to run `pod install` to remove the native dependencies

ishrak-imam (approved):

PR #1102, opened 2022-04-20
Bug-1100-app-crash

### LINKED ISSUE #1100: Bug: App crashes when clicked on democracy.
When clicked on democracy 

## Steps To Reproduce

1. Click on democracy on the network which has referendums
2. The app crashes.

https://user-images.githubusercontent.com/22363352/163836233-8b017631-f4e3-46bc-bb01-b75809297b6f.mp4

Ekaanth:
fixes #1100 

issue-1100

### DIFF
--- a/src/ui/components/ProposalCall.tsx
+++ b/src/ui/components/ProposalCall.tsx
@@ -55,14 +55,12 @@ function SubCall({subCall}: SubCallProps) {
   return (
     <Card mode="outlined" style={styles.subCallContainer}>
       <Card.Content>
-        <Text>
-          {subCall.meta && (
-            <>
-              <Paragraph>{subCall.meta}</Paragraph>
-              <Padder scale={0.5} />
-            </>
-          )}
-        </Text>
+        {subCall.meta ? (
+          <>
+            <Paragraph>{subCall.meta}</Paragraph>
+            <Padder scale={0.5} />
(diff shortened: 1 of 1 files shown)


src/ui/components/ProposalCall.tsx
```diff
@@ -55,12 +55,14 @@ function SubCall({subCall}: SubCallProps) {
...
     <Card mode="outlined" style={styles.subCallContainer}>
       <Card.Content>
-        {subCall.meta && (
-          <>
-            <Paragraph>{subCall.meta}</Paragraph>
-            <Padder scale={0.5} />
-          </>
-        )}
+        <Text>
+          {subCall.meta && (
```
silva-fj: This is not safe in react native, it is better to use `{subCall.meta ? <>...</> : null}`. I will add an eslint plugin to avoid this in a separate PR.

silva-fj: Yo can remove the `Text` wrapper

silva-fj (approved):
Well, it looks that there is a plugin for this but it does not seem to work.. 

We can do: `{something ? <Component /> : null}` or `{!!something && <Component />}`. Just be aware of this issue in react native, `0` or `''` or `NaN` will crash the app when using `&&`.

PR #1107, opened 2022-04-26
Issue 763 democracy summary teaser test case

### LINKED ISSUE #1137: Testcase: Democracy summary teaser test case

Ekaanth:
Democracy summary teaser test case


src/ui/components/DemocracySummaryTeaser.tsx
```diff
@@ -26,12 +26,20 @@ export function DemocracySummaryTeaser(props: Props) {
         <View style={styles.boxRow}>
           <Card mode="outlined" style={styles.card}>
             <View style={styles.itemRow}>
-              <StatInfoBlock title="Proposals">{String(data.activeProposals)}</StatInfoBlock>
-              <StatInfoBlock title="Total">{String(data.proposals)}</StatInfoBlock>
+              <StatInfoBlock title="Proposals" testID="activeProposals_id">
```
ishrak-imam: We should really limit our usage of attaching test ids into components. This is a rabbit whole we don't want to go into.
Should only be used as an escape hatch for truly unusual scenarios.

src/ui/components/DemocracySummaryTeaser.test.tsx
```diff
@@ -0,0 +1,882 @@
...
+    const activeProposal = getByTestId('activeProposals_id');
+    expect(within(activeProposal).getByText('0')).toBeTruthy();
+    const proposals = getByTestId('proposals_id');
+    expect(within(proposals).getByText('17')).toBeTruthy();
+    const referendums = getByTestId('referendums_id');
+    expect(within(referendums).getByText('60')).toBeTruthy();
+    const activeReferendums = getByTestId('activeReferendums_id');
+    expect(within(activeReferendums).getByText('1')).toBeTruthy();
+    const progressChart = getByTestId('progress_chart_id');
+    expect(within(progressChart).getByText('87% 3 days 10 hrs')).toBeTruthy();
```
ishrak-imam: Here we are getting the progress chart component with a test id and asserting a text such as `87% 3 days 10 hrs`.
Also we are taking a snapshot of the component. IMO this is redundant.

ishrak-imam (commented):
Not a big fan of snapshot testing. In reality it doesn't give us confidence that we want from writing tests.

Yes, for truly presentational components that possibly makes sense but we are not quite there yet with separation of components.

Writing useful tests also depends on better designing/dividing components. When we will be getting the UI/UX designs we should write components by keeping in mind about having useful tests also invest time in refactoring the current ones.

src/ui/components/DemocracySummaryTeaser.tsx
Ekaanth: Yes, I'm updating my test cases to not depend on`testID`

src/testUtils.tsx
```diff
@@ -5,6 +5,7 @@ import {ApolloProvider, ApolloClient, InMemoryCache, HttpLink} from '@apollo/cli
 import fetch from 'cross-fetch';
 
 import ThemeProvider from 'context/ThemeContext';
+import {create, TestRendererOptions} from 'react-test-renderer';
```
silva-fj: There should not be a case where we would need to use this api, please revert the changes in this file

silva-fj (approved):

ishrak-imam:
@Ekaanth Seems that the PR was merged with some failed checks. Can you please check?

PR #1126, opened 2022-05-02
Setting up firebase's remote config to test the token bridge on released builds

silva-fj:
With this PR we will be able to enable features on the app that are not ready for production in our released builds.

As initial setup the "ParachainApp" can be enabled remotely by adding an external address (I will add a notion page explaining how to enable this).

Also we have already in place the infrastructure to add analytics in the future.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -10,6 +10,11 @@ PODS:
     - React-Core (= 0.67.3)
     - React-jsi (= 0.67.3)
     - ReactCommon/turbomodule/core (= 0.67.3)
+  - Firebase/Analytics (8.15.0):
+    - Firebase/Core
+  - Firebase/Core (8.15.0):
+    - Firebase/CoreOnly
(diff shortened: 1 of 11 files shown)


ishrak-imam (approved):
LGTM!

Just a question: With this setup how can we enable the experiment for a group of addresses without adding or removing an address from firebase config one by one?

silva-fj:
> Just a question: With this setup how can we enable the experiment for a group of addresses without adding or removing an address from firebase config one by one?

If by "one by one" you mean adding additional config keys for each address, the answer is you don't have to.. As config value we are using a json string, so we can add as many addresses as we want in an array for example. Having said that, we don't need to add multiple addresses, the address is not to be used in any way to test the token bridge.. It's just an identifier to enable the screens.

This is not to be confused with experiments, we are not creating experiments with this

Ekaanth (approved):
LGTM 🎊

PR #1131, opened 2022-05-04
Fetch Tips from updated resolver

ishrak-imam:
### Description

Fetch tips from the updated resolver.


Ekaanth (approved):
LGTM

src/ui/screens/Tips/TipDetailScreen.tsx
```diff
@@ -39,7 +39,7 @@ function TipDetailContent({tip, toAccountDetails}: TipDetailProps) {
                 <Subheading>Finder</Subheading>
               </View>
               <View style={styles.addressContainer}>
-                <AccountTeaser account={tip.finder.account} onPress={() => toAccountDetails(tip.who.account.address)} />
+                <AccountTeaser account={tip.finder} onPress={() => toAccountDetails(tip.finder?.address as string)} />
```
silva-fj: Is the `as string` because the type can be `undefined`? If so, let's wrap the call with an if statement so we are sure this will never fail (even if we know the address should be defined)

silva-fj (approved):

PR #1134, opened 2022-05-05
Issue 1132 utility function test case

### LINKED ISSUE #1132: Testcase: utility functions, functionality testing.

Ekaanth:
fix #1132


src/utils.test.ts
```diff
@@ -0,0 +1,19 @@
+import {decimalKeypad, notEmpty} from './utils';
+
+test('utility function notEmpty', () => {
+  const testCase = ['', null, undefined, 'Test'];
+  const expected = [false, false, false, true];
```
silva-fj: Instead of this you can use [each](https://jestjs.io/docs/api#describeeachtablename-fn-timeout) to make tests easier to debug. If the test fails for a specific input you will know which one exactly

src/utils/address.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  const testCase = [POLKADOT_ADDRESS, KUSAMA_ADDRESS, LITMUS_ADDRESS];
+  const expected = ['1HDgY7…pKUKFw', 'D3KNzS…Sg6LXN', 'jcSKTf…pz8KvJ'];
+  testCase.forEach((address, i) => {
+    const decimalValue = toShortAddress(address);
+    expect(decimalValue).toBe(expected[i]);
+  });
+});
+
+test('isAddressValid function test', async () => {
+  const networkTypes = await waitFor(() => getAvailableNetworks());
```
silva-fj: `getAvailableNetworks` is a normal function, you are not rendering a component so you can just normal `async/await`. (`waitFor` is for testing components)

```suggestion
  const networkTypes = await getAvailableNetworks();
```

src/utils/balance.test.ts
```diff
@@ -0,0 +1,31 @@
+import {Registry} from 'src/api/hooks/useChainInfo';
+import {formattedStringToBn, stringToBn} from './balance';
+
+test('stringToBn test case', async () => {
```
silva-fj: If you are not awaiting the `async` keyword is not needed

src/utils/date.test.ts
```diff
@@ -0,0 +1,17 @@
...
+  expect(now).toBe('6 years ago');
+});
+
+test('getCurrentYear function', () => {
+  const fullYear = new Date().getFullYear();
+  const currentYear = getCurrentYear();
+  expect(currentYear).toEqual(fullYear);
+});
+
+test('fromNow function', () => {
```
silva-fj: is this a test for `fromNow` ?

src/utils/email.ts
```diff
@@ -1,6 +1,6 @@
 import {Linking} from 'react-native';
 
-type SendEmailProps = {
+export type SendEmailProps = {
```
silva-fj: NIT: Let's rename this type to `Options` or `SendEmailOptions`

src/utils/identity.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    isKnownGood: false,
+    isOutOfDate: false,
+    isLowQuality: false,
+  };
+  const expectedJudgement = {text: 'Erroneous', category: 'bad', icon: 'alert-circle-outline'};
+
+  const mapStatus = mapStatusText(isErroneousJudgement, false);
+  expect(mapStatus).toEqual(expectedJudgement);
+
+  const mapStatus_hasParent = mapStatusText(isErroneousJudgement, true);
```
silva-fj: let's not use `_`, always use camelCase

src/utils/identity.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  const isKnownGoodJudgement: IdentityJudgement = {
+    isErroneous: false,
+    isReasonable: true,
+    isKnownGood: false,
+    isOutOfDate: false,
+    isLowQuality: false,
+  };
+  const expectedJudgement = {text: 'Reasonable', category: 'good', icon: 'check-circle'};
+
+  const expectedJudgement_hasParent = {text: 'Reasonable', category: 'good', icon: 'link'};
```
silva-fj: same

src/utils/identity.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  const mapStatus = mapStatusText(isKnownGoodJudgement, false);
+  expect(mapStatus).toEqual(expectedJudgement);
+
+  const mapStatus_hasParent = mapStatusText(isKnownGoodJudgement, true);
```
silva-fj: same

src/utils/identity.test.ts
```diff
@@ -0,0 +1,95 @@
+import {SubstrateChainIdentityJudgement as IdentityJudgement} from 'src/generated/litentryGraphQLTypes';
+import {mapStatusText} from './identity';
+
+test('isErroneous test case', () => {
```
silva-fj, line 4: You can use [each](https://jestjs.io/docs/api#describeeachtablename-fn-timeout) for the entire file so you don't have to repeat the same for each data set

silva-fj (commented):
Thank you for adding all these tests Abhi, great work 👍🏼 

Just a few comments on things that can be improved.

src/utils/address.test.ts
silva-fj: I just noticed that `getAvailableNetworks` is not async, so no need to await

src/utils/identity.test.ts
Ekaanth: I'll keep this in mind, but in this case, we have different expected outputs for two other cases rewriting using `describe.each (...)` providing multiple objects will make it complicated.

src/utils/address.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  testCase.forEach((address, i) => {
+    const decimalValue = toShortAddress(address);
+    expect(decimalValue).toBe(expected[i]);
+  });
+});
+
+test('isAddressValid function test', () => {
+  const networkTypes = getAvailableNetworks();
+  const validAddressPolka = isAddressValid(networkTypes[0] as NetworkType, POLKADOT_ADDRESS);
+  expect(validAddressPolka).toBeTruthy();
```
silva-fj:
```suggestion
  expect(validAddressPolka).toBe(true);
```
In all these cases you should assert that it is `true`/`false`.. If we break the function like make it to return a `string` this test will not catch that

src/utils/email.ts
```diff
@@ -10,7 +10,17 @@ type SendEmailProps = {
   };
 };
 
-export async function sendEmail(props: SendEmailProps) {
+export async function sendEmail(props: SendEmailOptions) {
+  const url = composeEmailURL(props);
+  // check if we can use this link
```
silva-fj: Can you remove this comment?

silva-fj (approved):

PR #1153, opened 2022-05-09
Adding animated ProgressChart component

silva-fj:
This PR introduces a new `ProgressChart` component which is animated.

The old `Chart` component was only used in the Crowdloan screen so I only replaced it there. I would refactor the ProgressChartWidget component used in the other screens to have animations or create a better abstraction in a separate PR.

This is how it looks

https://user-images.githubusercontent.com/8374946/167426815-4b182422-5930-4808-927b-258ebdab39e9.mp4

While doing the refactoring I noticed that there are some performance issues in the `CrowdloanScreen` (that's why it looks a bit laggy in the Simulator), I will tackle it in a separate PR as well.


Ekaanth (approved):
This feature makes the graphs look alive!! 🎊👍🏾

ishrak-imam (approved):

PR #1162, opened 2022-05-11
Issue-1152 Improve loading states

ishrak-imam:
### Description

Introducing a Skeleton component to compose different loading UIs with more control on the shape.

Example:

```
<Skeleton width={50} /> // rectangular skeleton with 50% width
<Skeleton width={50} circle /> // circle with radius 50px
<Skeleton width={50} height={20} /> // rectangular skeleton with 50% width and height 20px
```

https://user-images.githubusercontent.com/16683923/167829472-316ceb90-3d4d-43e9-81d6-1c12f1d56d29.mp4

https://user-images.githubusercontent.com/16683923/167829505-41eb41c4-0065-4565-9b23-4a10d1a3b557.mp4

https://user-images.githubusercontent.com/16683923/167829518-1ee71798-0097-47ed-826d-cfe5cdbdb5b9.mp4


Ekaanth (approved):
The new loading state looks good!! 🎊

src/ui/library/Skeleton.tsx
```diff
@@ -0,0 +1,57 @@
...
+};
+
+const AnimatedLG = Animated.createAnimatedComponent(LinearGradient);
+
+export function Skeleton({width, height, circle = false}: Props) {
+  const widthValue = !circle ? `${width}%` : width;
+  const heightValue = height ? height : !circle && !height ? 25 : width;
+  const borderRadius = circle ? Math.floor(width / 2) : 4;
+
+  const animatedValue = new Animated.Value(0);
```
silva-fj: Let's put this in a ref to add it as a dependency in the effect

src/ui/library/Skeleton.tsx
```diff
@@ -0,0 +1,57 @@
...
+  React.useEffect(() => {
+    Animated.loop(
+      Animated.timing(animatedValue, {
+        toValue: 1,
+        duration: 2000,
```
silva-fj: Can you try decreasing the duration a bit?

src/ui/library/Skeleton.tsx
```diff
@@ -0,0 +1,57 @@
...
+        easing: Easing.linear,
+        useNativeDriver: true,
+      }),
+    ).start();
+  });
+
+  const translateX = animatedValue.interpolate({
+    inputRange: [0, 1],
+    outputRange: [circle ? -width : -250, circle ? width : 250],
```
silva-fj: You could also calculate the output with the width or screen dimension so it can fit good in bigger screens as well if the loading state placeholder should take more horizontal space

silva-fj (approved):
Awesome, now that we have the `Skeleton` component we can build loading UI's that resemble the real UI. Once we have the new designs we can start building those.

src/ui/library/Skeleton.tsx
ishrak-imam: Actually the the recordings were before adjusting the values a little bit, Duration was higher. Duration 2000 seems to be ok. Otherwise it is too fast.

src/ui/library/Skeleton.tsx
ishrak-imam: Since the width becomes a percentage in case of a rectangular skeleton we need to calculate the actual width onLayout and this makes the animation a bit shaky. I did implemented the screen width calculation and removed it. Will put it back.

PR #1168, opened 2022-05-12
Refactoring TxContext

silva-fj:
This is a small clean up to the code related to TxContext. 

Relevant changes:
- All Tx related components are grouped in the Tx module.
- Modalize component replaced with the new BottomSheet component. 

Later when/if we integrate other ways to sign transactions (like WalletConnect v2) we can refactor this to separate concerns and make it easier to modify.


src/ui/library/BottomSheet.tsx
```diff
@@ -31,8 +31,9 @@ export function useBottomSheet() {
   );
 
   const BottomSheet = React.useCallback(
-    ({children}: BottomSheetProps) => (
+    ({children, ...props}: BottomSheetProps) => (
```
ishrak-imam, line 34: Will be needing this in my PR.

ishrak-imam, line 34: Will be needing this change in my PR 👍

ishrak-imam, line 34: Will be needing this change in my PR 👍

ishrak-imam (approved):
Changes look good.

PR #1171, opened 2022-05-13
Refactor Modalize with BottomSheet

ishrak-imam:
### Description

Since we now have a new implementation of a BottomSheet component we don't need the Modalize any more. This PR mainly focuses on replacing the Modalize instances with BottomSheet.

Additionally, because of the implementation detail of the Modalize component we were using Navigation screens to display the contents in Modalize which we don't need any more. This PR also changes those. Some Screens now live as independent components inside `@ui/components` path and used as contents of the BottomSheet.


ishrak-imam:
Although the PR is open for review, these changes demand a thorough testing. I'll keep doing it mean while reviews are welcome.

src/ui/library/BottomSheet.tsx
```diff
@@ -43,10 +53,10 @@ export function useBottomSheet() {
         contentHeight={animatedContentHeight}
         enablePanDownToClose={true}
         animateOnMount={true}>
-        <BottomSheetView onLayout={handleContentLayout}>{children}</BottomSheetView>
+        <BottomSheetView onLayout={handleContentLayout}>{children ? children : noContent}</BottomSheetView>
```
ishrak-imam: We need this because passing `undefined` or `null` as children to the BottomSheet makes it behave unexpectedly. This is just to handle and edge case in the [dynamic content generation for BottomSheet](https://github.com/litentry/litentry-app/blob/effddf0dfc3146afca243f9388b254864bf0d5e3/src/ui/screens/AccountsScreen.tsx#L66), but ideally the BottomSheet will never be opened without any content.

src/ui/components/Account/AccountBalance.tsx
```diff
@@ -0,0 +1,52 @@
...
+          <ItemRight>
+            <Caption>{balance.formattedFreeFrozen}</Caption>
+          </ItemRight>
+        )}
+      />
+    </>
+  );
+}
+
+function ItemRight({children}: {children: React.ReactNode}) {
```
silva-fj: Do we need this container? We could probably just do `textAlign` center in the `Caption`

src/ui/library/BottomSheet.tsx
ishrak-imam: An updated implementation of [dynamic BottomSheet](https://github.com/litentry/litentry-app/blob/effddf0dfc3146afca243f9388b254864bf0d5e3/src/ui/screens/AccountsScreen.tsx#L66) is [here](https://github.com/litentry/litentry-app/pull/1179)

But we still need to check if the children is null.

jest/setup.js
```diff
@@ -1,3 +1,4 @@
+import React from 'react';
```
silva-fj, line 1: Why do you need to import `React` here?

mocks/handlers/bountyHandler.js
```diff
@@ -54,8 +54,8 @@ export const bountyHandler = graphql.query('getBounty', (_, res, ctx) => {
               formattedReserved: '40.0740 DOT',
               free: '2419260000000',
               formattedFree: '241.9260 DOT',
-              freeFrozen: '0',
```
silva-fj, line 57: Do you need to rebase?

src/ui/components/AddBounty.tsx
```diff
@@ -0,0 +1,161 @@
...
+import {useBountiesSummary} from 'src/api/hooks/useBountiesSummary';
+import type {BN} from '@polkadot/util';
+import {formattedStringToBn} from 'src/utils/balance';
+import {countUtf8Bytes} from 'src/utils';
+
+type Props = {
+  onClose: () => void;
+};
+
+export function AddBounty({onClose}: Props) {
```
silva-fj, line 24: Is it the same UI? Can you add some screenshots for the screens that are different (if any)

src/ui/components/IdentityGuide.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <GuideRow
+            text="6. Once your identity has been set you should see a successful message"
+            image={require('src/image/identity_guide/step6.jpg')}
+            direction="row"
+          />
+        </View>
+      </PagerView>
+      <Padder scale={1} />
+      <View style={styles.footer}>
+        <Button mode="outlined" onPress={onClose}>{`Skip`}</Button>
```
silva-fj:
```suggestion
        <Button mode="outlined" onPress={onSkip}>{`Skip`}</Button>
```

src/ui/components/SendFund.tsx
```diff
@@ -0,0 +1,124 @@
...
+import {useSnackbar} from 'context/SnackbarContext';
+import {BN_ZERO} from '@polkadot/util';
+import {useFormatBalance} from 'src/hooks/useFormatBalance';
+import {stringToBn as stringToBnUtil, formattedStringToBn} from 'src/utils/balance';
+import AddressInput from '@ui/components/AddressInput';
+import {Layout} from '@ui/components/Layout';
+
+type Props = {
+  address: string;
+  onClose: () => void;
```
silva-fj: NIT: 
```suggestion
  onFundsSent: () => void;
```

src/ui/library/BottomSheet.tsx
```diff
@@ -44,10 +65,21 @@ export function useBottomSheet() {
         enablePanDownToClose={true}
         animateOnMount={true}
         {...props}>
-        <BottomSheetView onLayout={handleContentLayout}>{children}</BottomSheetView>
+        <BottomSheetView onLayout={handleContentLayout}>
+          {children ? children : bottomSheetContent ? bottomSheetContent : noContent}
```
silva-fj, line 69: You can use typescript to force the user to always pass a children or the content to openBottomSheet so we don't have to handle the `noContent` case, it would also prevent the developers to forget passing any.. We should never display `No content`

src/ui/screens/MotionDetailScreen.tsx
```diff
@@ -104,7 +102,7 @@ export function MotionDetailScreen({route, navigation}: PropTypes) {
         <View style={globalStyles.spaceBetweenRowContainer}>
           <Card style={[styles.item, styles.left]}>
             <Card.Content>
-              <StatInfoBlock title="Section">{_.capitalize(proposal.section)}</StatInfoBlock>
+              <StatInfoBlock title="Section">{capitalize(proposal.section)}</StatInfoBlock>
```
silva-fj: You can also apply styles: `{textTransform: 'uppercase'}`

src/utils.ts
```diff
@@ -11,3 +11,7 @@ export function decimalKeypad(numericValue: string): string {
   const number = numericValue.replace(',', '.');
   return number.split('.').length <= 2 ? number : number.slice(0, -1);
 }
+
+export function countUtf8Bytes(str: string): number {
```
silva-fj, line 15: Can you add a test for this? Or create an issue to add it

silva-fj (commented):
Thank you for cleaning this up, great work!

src/ui/components/Account/AccountBalance.tsx
ishrak-imam: We need the vertical alignment here. We could use `textAlignVertical` but that only works on android. To make it work in iOS need to use a `lineHeight` with that.

jest/setup.js
ishrak-imam: For this `jest.mock('react-native-qrcode-scanner', () => <></>);` we need React in the scope. But once we refactor the QR scanner component will not need this anymore.

mocks/handlers/bountyHandler.js
ishrak-imam: No, some values were not cleaned-up.

src/ui/components/AddBounty.tsx
ishrak-imam: It's the same UI. Just the component was relocated.

src/utils.ts
ishrak-imam: https://github.com/litentry/litentry-app/issues/1186

silva-fj (approved):

PR #1179, opened 2022-05-19
useDynamicBottomSheet hook

ishrak-imam:
### Description

In cases like where we need to use multiple bottomsheet in a single screen this `useDynamicBottomSheet` is abstracting the logic of displaying the correct content according to the need.


Ekaanth (approved):
LGTM!!

src/hooks/useDynamicBottomSheet.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import {useBottomSheet, BottomSheetProps} from '@ui/library';
+
+type DynamicBottomSheetProps = Omit<BottomSheetProps, 'children'>;
+
+export function useDynamicBottomSheet() {
```
silva-fj: What I had in mind was to implement this in the `useBottomSheet` hook so we have a single component.

The logic could be the following:

```
...
 <BottomSheetView>{bottomSheetContent ?? children}</BottomSheetView>
...

function openBottomSheet(content?: React.ReactNode) {
   if (content) {
     setBottomSheetContent(content);
   }
   bottomSheetRef.current?.expand();
}
```

or we could expose another function like `updateBottomSheetContent(content)`

silva-fj (commented):
Awesome idea, Let's implement that in the current hook so we have a single component 👍🏼

src/ui/components/ReceiveFund.tsx
```diff
@@ -14,9 +14,17 @@ type Props = {
...
 
+const QR_CODE_DIMENSION = Dimensions.get('screen').width * 0.5;
+
 export function ReceiveFund({address, onClose}: Props) {
   const snackbar = useSnackbar();
-  const [imageUri] = useState(() => getAccountQRCode(address));
+  const [imageUri, setImageUri] = useState('');
+
+  useEffect(() => {
+    setTimeout(() => {
```
silva-fj, line 24: Why do we need `setTimeout` here?

src/ui/library/BottomSheet.tsx
```diff
@@ -9,22 +9,33 @@ import RNBottomSheet, {
...
     useBottomSheetDynamicSnapPoints(initialSnapPoints);
 
-  const openBottomSheet = () => {
-    bottomSheetRef.current?.expand();
-  };
+  const openBottomSheet = React.useCallback((content?: React.ReactNode) => {
+    if (content) {
+      setBottomSheetContent(content);
+    }
+    setTimeout(() => {
```
silva-fj, line 28: Same here, why do we need this?

silva-fj (approved):

PR #1208, opened 2022-05-31
Fixing Drawer bug

### LINKED ISSUE #1058: Bug: Investigate drawer(s) not responding

ishrak-imam:
Fixes #1058


silva-fj (approved):
I think we could also listen for the event dispatched and close the drawer there (It would have been less changes). But if this fixes the issue it's also fine 👍🏼

PR #1226, opened 2022-06-03
Fixing auction screen UI glitch

ishrak-imam:
### Description

Before:

After:


src/ui/screens/Parachains/AuctionsScreen.tsx
```diff
@@ -36,6 +38,7 @@ export function AuctionsScreen() {
 
   const {auctionsInfo, latestAuction} = auction;
   const {winningBid, leasePeriod, endingPeriod, raised, raisedPercent} = latestAuction;
+  const formattedRaised = formatBalance(raised, {isShort: true});
```
silva-fj, line 41: We will get this from the api now, right?

silva-fj (approved):

src/ui/screens/Parachains/AuctionsScreen.tsx
ishrak-imam: Yes we get a `formattedRaised` value from the api but in this case that's not it. We get the longer version from the api and here we are formatting the actual value into a shorter one.

PR #1237, opened 2022-06-10
1232 test case accounts

### LINKED ISSUE #1232: Test case: Accounts

Ekaanth:
---
  - https://app.asana.com/0/0/1202804937385427


src/ui/components/Account/AccountRegistration.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+} as SubstrateChainAccountRegistration;
+
+const accountInfo = [
+  ['https://twitter.com/@purestakeco', '@purestakeco'],
+  ['https://matrix.to/#/@purestakeco:matrix.parity.io', '@purestakeco:matrix.parity.io'],
+  ['https://www.purestake.com/', 'https://www.purestake.com/'],
+];
+
+describe('AccountRegistration', () => {
+  it('should render the AccountRegistration component with data', async () => {
```
silva-fj: > should render the AccountRegistration component with data

What data? You should make assertions here on what the user should see otherwise the test does not add any value

src/ui/components/Account/AccountTeaser.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+  },
+} as unknown as SubstrateChainAccount;
+
+const accountNavigation = {
+  navigate: jest.fn(),
+};
+
+describe('AddressInfoPreview', () => {
+  it('should render the AccountTeaser component', async () => {
+    const {findByText} = render(<AccountTeaser account={account} onPress={accountNavigation.navigate} />);
```
silva-fj, line 35: This test is exactly the same as the one in `src/ui/components/Account/Account.test.tsx` Are there any differences in the components for the users?

Ekaanth: Nope, the same test case gets repeated several times but on different screens. We are providing other same test cases on various screens.

PR #1245, opened 2022-06-11
1244 test case create account

### LINKED ISSUE #1244: test case: create account

Ekaanth:


src/ui/screens/CreateAccount/CreateAccountScreen.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+  goBack: () => jest.fn,
+} as unknown as NavigationProp<AccountsStackParamList>;
+
+const route = {
+  params: {
+    mnemonic: 'west bar upon arena all remove return era local spoon edge use',
+  },
+} as RouteProp<AccountsStackParamList, typeof createAccountScreen>;
+
+const mockCreateAddressFromMnemonic = jest.fn(() => Promise.resolve());
```
silva-fj: This mock should return an address and if we display it in the screen you should assert that is displayed

src/ui/screens/CreateAccount/CreateAccountScreen.tsx
```diff
@@ -114,17 +117,18 @@ export function CreateAccountScreen({
             autoComplete="off"
             error={Boolean(account.password) && passwordStrength < 3}
           />
-          <HelperText type="error" visible={passwordError}>
+          <HelperText type="error" visible={passwordError} testID="weak-password">
```
silva-fj: Why do you need the testID? The user should be able to see the text right?

PR #1257, opened 2022-06-15
1251 testcase sub-identities

### LINKED ISSUE #1251: testcase: subidentities

Ekaanth:


mocks/handlers/accountHandler.js
```diff
@@ -48,6 +48,7 @@ export const accountHandler = graphql.query('getAccount', (req, res, ctx) => {
           feeFrozen: '198480993716691',
           formattedFeeFrozen: '19,848.0993 DOT',
         },
+        subAccounts: [],
```
ishrak-imam: Don't we need any mock data for the sub accounts??

ishrak-imam (approved):

src/ui/screens/SubIdentities/AddSubIdentity.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const bottomSheetModal = {
+  dismiss: jest.fn(),
+};
+
+const addSubIdentity = {
+  subIdentity: jest.fn(),
+};
+
+describe('', () => {
```
silva-fj: Let's add the description

src/ui/screens/SubIdentities/AddSubIdentity.test.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import {render, fireEvent} from 'src/testUtils';
+import {AddSubIdentity} from './AddSubIdentity';
+
+jest.useFakeTimers();
```
silva-fj: Why do you need this?

src/ui/screens/SubIdentities/AddSubIdentity.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+    await findByText('Add Identity');
+  });
+
+  it('should close the bottom-sheets when pressed on cancel button', async () => {
+    const bottomSheetsSpy = jest.spyOn(bottomSheetModal, 'dismiss');
+    const {findByText, findByTestId} = render(
+      <AddSubIdentity onClose={bottomSheetModal.dismiss} onAddPress={addSubIdentity.subIdentity} />,
+    );
+    await findByText('Cancel');
+    fireEvent.press(await findByTestId('cancel-identity-button'));
```
silva-fj: In this case (finding the button to fire a press event) it is better to get it by role or text, the `testID` is needed when you want to assert that the button is disabled or not.

src/ui/screens/SubIdentities/AddSubIdentity.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+    expect(bottomSheetsSpy).toBeCalledTimes(1);
+  });
+
+  it('should add new sub-identity', async () => {
```
silva-fj: How about:
` it('should call onAddPress with the right arguments when the form is valid and Add Identity button is pressed', () => {...})`

We are not actually testing that the sub-identity was added to something

src/ui/screens/SubIdentities/RegisterSubIdentitiesScreen.test.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import {NavigationProp, RouteProp} from '@react-navigation/native';
+import {AccountsStackParamList} from '@ui/navigation/navigation';
+import {render} from 'src/testUtils';
+import {RegisterSubIdentitiesScreen} from './RegisterSubIdentitiesScreen';
+import {registerSubIdentitiesScreen} from '@ui/navigation/routeKeys';
+
+jest.useFakeTimers();
```
silva-fj: Same, why do you need to add this?

src/ui/screens/SubIdentities/AddSubIdentity.test.tsx
Ekaanth: Initially, when I was writing a test case using the `test` function, I had an import teardown, so adding this resolved it. Now I'm testing within describe, so I don't need it. It was removed in the latest push.

src/ui/screens/SubIdentities/AddSubIdentity.test.tsx
Ekaanth: > We are not testing that the sub-identity was added to something

In this component, we don't have the actual implementation. While this component accepts the `address` and sends an object, that's what I'm mocking in this.

mocks/handlers/accountHandler.js
Ekaanth: I have tried to mock the data, but I think the data I mocked is wrong, and unable to find a relevant example account. Is it possible for you to provide a mock for this?

ishrak-imam:
```
{
  substrateChainAccount(address: "13KJ3t8w1CKMkXCmZ6s3VwdWo4h747kXE88ZNh6rCBTvojmM") {
    address
    subAccounts {
      address
    }
  }
}
```

This account has sub accounts. You can get the mock from here. Add it as an external account in the app and you can get the data.

silva-fj (approved):

PR #1267, opened 2022-06-21
Adding testing library eslint plugin

silva-fj:
This plugin helps to catch incorrect ways to setup tests (specially useful for async tests)


Ekaanth (approved):
This was much needed 👍🏾

PR #1236, opened 2022-06-09
WebView implementation of the wallet functionalities

ishrak-imam:
Because of not having the WASM support in react-native environment we were using [react-native-substrate-sign](https://github.com/paritytech/react-native-substrate-sign) for wallet related functionalities. But that project is not maintained anymore and it also doesn't support the address formatting of the parachain networks.

That's why we need to bring back one of the other [options](https://github.com/litentry/litentry-app/pull/553) that was considered before, when we faced this performance issue while using polkadot keyring.

The idea is to load the polkadot keyring related packages inside a WebView and interact with that from the app.

Following are the functionalities that we currently need 

- [x] Generate mnemonic
- [x] Create account
- [x] Add external account
- [x] Remove account
- [x] Import account from seed
- [x] Import account from JSON
- [x] Export account to JSON file
- [x] Sign transactions
- [x] Test on real devices 
- [x] Clean up the code

PR #1274, opened 2022-06-22
Cleaning up after WebView implementation

ishrak-imam:
### Description

Since we are moving towards the [WebView implementation of wallet features](https://github.com/litentry/litentry-app/pull/1236) we have some clean-up opportunities in the code.

For example removing the `react-native-substrate-sign` library and all of its implementations is one of them. Additionally since we are managing application state using [recoil](https://recoiljs.org/) now, we can get rid of the Accounts context provider too. And since because of an implementation detail of moving the selected network state into a persisted recoil atom we can actually get rid of the Network context provider as well.

This PR mainly takes care of these clean-ups.

- [x] Cleaning up react-native-substrate-sign
- [x] Removing Accounts context provider
- [x] Moving active network state into recoil atom
- [x] Removing Network context provider
- [x] Changing firebase remote config initialisation process


geastwood:
This is good effort, however can we also have a clear task list to provide better overview on what we trying to achieve within this PR?

src/polkadotApi/useAppAccounts.ts
```diff
@@ -1,16 +1,19 @@
-import {useMemo} from 'react';
-import {useNetwork} from 'context/NetworkContext';
-import {useRecoilState} from 'recoil';
+import {networkState} from '@atoms/network';
+import {selector, useRecoilState, useRecoilValue} from 'recoil';
 import {appAccountsState} from './atoms';
 
+const netWorkAccountState = selector({
```
silva-fj:
```suggestion
const networkAccountsState = selector({
```

src/ui/screens/Council/MotionsScreen.tsx
```diff
@@ -16,14 +16,14 @@ import {SelectAccount} from '@ui/components/SelectAccount';
...
-import type {Account as LocalAccount} from 'context/AccountsContext';
-import {useNetwork} from 'context/NetworkContext';
+import {useNetwork} from '@atoms/network';
 import type {SupportedNetworkType} from 'src/types';
 import {Caption, Card, Divider} from 'react-native-paper';
 import {ProposalCall} from '@ui/components/ProposalCall';
 import {ItemRowBlock} from '@ui/components/ItemRowBlock';
 import {AccountTeaser} from '@ui/components/Account/AccountTeaser';
 import {getProposalTitle} from 'src/utils/proposal';
+import type {Account as LocalAccount} from '@polkadotApi/types';
```
silva-fj: We should probably make the types a bit more consistent, we use aliases for `Account`, `AppAccount` and `LocalAccount`. But we can tackle this in a separate issue

src/ui/screens/DevScreen.tsx
```diff
@@ -5,11 +5,11 @@ import {InAppNotificationContent, InAppNotificationContext} from 'context/InAppN
 import {useApi} from 'context/ChainApiContext';
 import {ScrollView} from 'react-native-gesture-handler';
 import globalStyles, {standardPadding} from '@ui/styles';
-import {useNetwork} from 'context/NetworkContext';
+import {useNetwork} from '@atoms/network';
 import SafeView, {noTopEdges} from '@ui/components/SafeView';
 import {useConvictions} from 'src/api/hooks/useConvictions';
 import {SelectAccount} from '@ui/components/SelectAccount';
-import {useAccounts} from 'context/AccountsContext';
+import {useActiveAccount} from 'src/atoms/activeAccount';
```
silva-fj:
```suggestion
import {useActiveAccount} from '@atoms/activeAccount';
```

silva-fj (approved):

PR #1288, opened 2022-06-29
Issue-1268 Changing  push notification implementation

ishrak-imam:
### Description

Until now we were using react-query to manage different states of the push notification infrastructure. Since, we are planning to completely remove react-query as a dependency (data fetching is now handled by apollo-client) the implementation of the push notification infrastructure is changed to using recoil instead. Also different push notification related logics are now isolated now.


src/atoms/pushNotification.ts
```diff
@@ -0,0 +1,200 @@
...
+    [subscriptionTopics, setSubscriptionTopics],
+  );
+
+  const subscribeToAllTopics = useCallback(() => {
+    const ids = Object.keys(subscriptionTopics) as SUBSCRIPTION_ID[];
+    const subscriptionPromises = ids.map((topicId) => messaging().subscribeToTopic(topicId));
+
+    Promise.all(subscriptionPromises);
+
+    const subscribedTopics = ids.reduce((subscription, topicId) => {
```
silva-fj: We probably should not assume what all requests to subscribeToTopic will succeed, as they can fail for multiple reasons (e.g: token has expired or network issue). It will be safer to await for `Promise.allSettled` and mark them as true for each fulfilled promise.

Consider the same for single topic subscriptions

src/atoms/pushNotification.ts
```diff
@@ -0,0 +1,200 @@
...
+    }, {} as Subscription);
+
+    setSubscriptionTopics(subscribedTopics);
+  }, [subscriptionTopics, setSubscriptionTopics]);
+
+  const unSubscribeToAllTopics = useCallback(() => {
+    const ids = Object.keys(subscriptionTopics) as SUBSCRIPTION_ID[];
+    const subscriptionPromises = ids.map((topicId) => messaging().subscribeToTopic(topicId));
+
+    Promise.all(subscriptionPromises);
```
silva-fj: Same here

src/ui/screens/PermissionGrantingPrompt.tsx
```diff
@@ -28,27 +26,12 @@ function PermissionGrantingPromptScreen({skipPnPermission, allowSkip = true}: Pr
           <Caption style={globalStyles.textCenter}>
             Enable notifications to make sure you don't miss out on important events
           </Caption>
-          {error ? (
```
silva-fj, line 31: Don't we want to handle errors?

src/utils/firstAppStart.ts
```diff
@@ -0,0 +1,17 @@
+import * as Storage from 'service/PersistedObjectStorage';
+
+const IS_APP_STARTED_ONCE = 'is_app_started_once';
```
silva-fj: NIT: 
```suggestion
const HAS_APP_STARTED_ONCE = 'has_app_started_once';
```
```suggestion
const IS_APP_STARTED_ONCE = 'is_app_started_once';
```

src/ui/screens/PermissionGrantingPrompt.tsx
ishrak-imam: In this case it is prompting the native notification permission dialog. User can either `allow` or `don't allow` and both of the states are reflected accordingly in the UI.

Permission "Don't allow":
https://user-images.githubusercontent.com/16683923/176664107-e8939120-c804-4696-80a3-48d354c0fbd8.mp4

Permission "Allow":
https://user-images.githubusercontent.com/16683923/176664771-392d38e4-d5eb-47b7-b82b-451e083e4a30.mp4

silva-fj (approved):

Ekaanth (approved):

PR #1305, opened 2022-07-13
Migrating polkadot's api usage to the WebView implementation

silva-fj:
In this PR we have migrated all usages of the polkadot libraries (the ones that use crypto functions) to the webview implementation.

The API usage for making transactions has not changed much, but it has been improved, the relevant changes are:
- `startTx` from `useStartTx` have types to make sure we are passing the right params to the `tx` methods.
- The crypto util functions are now asynchronous.
- We have re-implemented Polkadot's Identicon component (thanks to @ishrak-imam).
- The litentry rococo network has been integrated so we could test transactions with fake tokens (thanks to @ishrak-imam)

With these changes we can see a major performance improvement in the app, not to mention that the app should not crash anymore due to websocket issues


ishrak-imam:
Working on fixing the failed tests...

ishrak-imam (approved):
LGTM!!

PR #1307, opened 2022-07-14
Creating custom Identicon component

ishrak-imam:
### Description

Since we are moving all the polkadot api related functionality into the WebView, we had to provide our own implementation of the polkadot identicon component, since it uses some crypto functions to generate circle values.

Here we create a custom Identicon component that can use async crypto functions and replace the usage with the new one.


src/ui/library/Identicon.tsx
```diff
@@ -0,0 +1,196 @@
+import React from 'react';
```
silva-fj: Let's move this file to `ui/components`, components in the library should only have styles or depend on the base component libraries (like react-native-paper)

silva-fj (approved):
Great job 👍🏼

PR #1314, opened 2022-07-27
Issue 1173 democracy from indexer

ishrak-imam:
### Description

In this PR we fetch the proposal data from the indexer. Now we have the title and description of the proposals. In addition the proposal teasers are refactored with new data and proposal call information is removed from motion proposals.


src/context/LitentryApiContext.tsx
```diff
@@ -10,6 +10,17 @@ type LitentryApiContextType = {
...
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const mergeFunc = (existing: any, incoming: any, {args}: any) => {
+  const offset = args ? args.offset : 0;
+  const merged = existing ? existing.slice(0) : [];
+  for (let i = 0; i < incoming.length; ++i) {
+    merged[offset + i] = incoming[i];
+  }
+  return merged;
+};
+/* eslint-enable @typescript-eslint/no-explicit-any */
```
silva-fj, line 22: Why do we need this one?

src/context/LitentryApiContext.tsx
```diff
@@ -10,6 +10,17 @@ type LitentryApiContextType = {
 
 const LitentryApiContext = React.createContext<LitentryApiContextType | undefined>(undefined);
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const mergeFunc = (existing: any, incoming: any, {args}: any) => {
```
silva-fj, line 14: I wish we could be more explicit with the types

silva-fj (approved):

src/context/LitentryApiContext.tsx
ishrak-imam: disabling eslint for this code block only https://github.com/litentry/litentry-app/blob/3f25355a16bee1c9ef1a8947f7ee2f96783405df/src/context/LitentryApiContext.tsx#L13-L22

src/context/LitentryApiContext.tsx
ishrak-imam: We can investigate this further as a future work.

PR #1319, opened 2022-08-01
reordered the providers.

### LINKED ISSUE #1275: Multiple instance of Apollo client created.
With the latest test case implementation, we encountered an issue with the apollo client. However, it was resolved using the `LitentryApiClientProvider` instance, thus creating multiple models of the apollo client in the application.

Ekaanth:
fixed #1275


src/context/ThemeContext.tsx
```diff
@@ -3,6 +3,8 @@ import {Platform, StatusBar} from 'react-native';
 import {usePersistedState} from '@hooks/usePersistedState';
 import {Provider as PaperProvider} from '@ui/library';
 import {themeDark, themeLight} from '@ui/library/theme';
+import {LitentryApiClientProvider} from './LitentryApiContext';
```
silva-fj: These Providers don't belong here, they have nothing to do with the Theme

src/polkadotApi/PolkadotApiWebView.tsx
```diff
@@ -345,7 +345,6 @@ function useWebViewOnMessage(resolversRef: ResolversRef, postMessage: PostMessag
 
   const webViewOnMessage = React.useCallback(
     (event: WebViewMessageEvent) => {
-      console.info('WebView Response: ', event.nativeEvent.data);
```
silva-fj, line 348: You can wrap this into `(__DEV__) {...}` but let's keep it. It's quite helpful when developing

silva-fj (commented):
What are you trying to fix with this PR?

Ekaanth:
> What are you trying to fix with this PR?

When the app is running, we create 2 instances of `LitentryApiClientProvider`, one for the testcase and the other for the application; I aim to create a single instance.

silva-fj:
> > What are you trying to fix with this PR?
> 
> When the app is running, we create 2 instances of `LitentryApiClientProvider`, one for the testcase and the other for the application; I aim to create a single instance.

When the app is running the test environment does not run.. I'm not sure I understand the problem you are trying to solve

ishrak-imam:
The issue is that, `LitentryApiClientProvider` is already instantiating an apollo client. In the test environment we are creating another one and calling the `client.clearStore()` from that. Which is in fact not useful since the test environment is using the apollo client that is being created in the `LitentryApiClientProvider`

Ideally what we want it to use the same instance of the client in the test environment.

silva-fj:
> The issue is that, `LitentryApiClientProvider` is already instantiating an apollo client. In the test environment we are creating another one and calling the `client.clearStore()` from that. Which is in fact not useful since the test environment is using the apollo client that is being created in the `LitentryApiClientProvider`
> 
> Ideally what we want it to use the same instance of the client in the test environment.

Yeah I remember the issue, what I don't get is what this PR is for (It is not addressing the issue with the apollo client)

ishrak-imam (approved):
Let's have this merged. We don't need to publish the app for this so please add skip CI.

PR #1332, opened 2022-08-04
flashList Installed

### LINKED ISSUE #1326: Replace FlatList with FlashList on long lists
FlashList is a new library develop and maintained by Shopify that has performance improvements.

https://shopify.github.io/flash-list/

Ekaanth:
fixed #1326

### DIFF
--- a/src/ui/components/Account/AccountsList.tsx
+++ b/src/ui/components/Account/AccountsList.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {FlatList, View} from 'react-native';
+import {View} from 'react-native';
 import {Divider} from '@ui/library';
 import type {AccountInfo} from 'src/api/hooks/useAccount';
 import {NavigationProp, useNavigation} from '@react-navigation/native';
--- a/src/ui/components/Account/AccountsList.tsx
+++ b/src/ui/components/Account/AccountsList.tsx
@@ -1,13 +1,12 @@
 import React from 'react';
 import {View} from 'react-native';
-import {Divider} from '@ui/library';
+import {Divider, FlatList} from '@ui/library';
 import type {AccountInfo} from 'src/api/hooks/useAccount';
 import {NavigationProp, useNavigation} from '@react-navigation/native';
 import {AppStackParamList} from '@ui/navigation/navigation';
 import {accountScreen} from '@ui/navigation/routeKeys';
 import {AccountTeaser} from './AccountTeaser';
 import globalStyles from '@ui/styles';
-import {FlashList} from '@shopify/flash-list';
 
 type Props = {
   accounts: AccountInfo[];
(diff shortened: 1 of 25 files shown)


src/ui/screens/Democracy/DemocracyScreen.tsx
```diff
@@ -244,7 +247,7 @@ function Row({label, children}: {label: string; children: React.ReactNode}) {
 
 const styles = StyleSheet.create({
   listContainer: {
-    padding: standardPadding * 2,
+    paddingVertical: standardPadding * 2,
```
ishrak-imam, line 249: Seems like you are just trying to reduce the horizontal padding. That can also be done like following.

```suggestion
    paddingVertical: standardPadding * 2,
    paddingHorizontal: standardPadding,
```

src/ui/screens/Democracy/DemocracyScreen.tsx
```diff
@@ -261,6 +264,7 @@ const styles = StyleSheet.create({
   },
   standardPadding: {
     padding: standardPadding,
+    marginHorizontal: standardPadding,
```
ishrak-imam, line 266: And removing this.

```suggestion
```

src/ui/screens/Democracy/DemocracyScreen.tsx
Ekaanth: The reason I'm doing something like this is that I get a warning as :
image

src/ui/screens/Democracy/DemocracyScreen.tsx
Ekaanth: We need this so that we can have the margin on both sides else the card will be stuck to the end of the screen like this 
image

src/ui/screens/Democracy/DemocracyScreen.tsx
ishrak-imam: I see, what you mean.

src/ui/screens/Democracy/DemocracyScreen.tsx
ishrak-imam: In that case change the name `standardPadding` to `itemPadding` or something like that.

Ekaanth: Renamed to `teaserCard` since I use it in both Referendums and Proposals teasers

src/ui/components/Account/AccountsList.tsx
```diff
@@ -1,12 +1,13 @@
...
-import {FlatList, View} from 'react-native';
+import {View} from 'react-native';
 import {Divider} from '@ui/library';
 import type {AccountInfo} from 'src/api/hooks/useAccount';
 import {NavigationProp, useNavigation} from '@react-navigation/native';
 import {AppStackParamList} from '@ui/navigation/navigation';
 import {accountScreen} from '@ui/navigation/routeKeys';
 import {AccountTeaser} from './AccountTeaser';
 import globalStyles from '@ui/styles';
+import {FlashList} from '@shopify/flash-list';
```
silva-fj: let's export this from the @library, it will make it easier later if we create an abstraction for lists

ishrak-imam (approved):
Please test on both platforms before merging.

PR #1337, opened 2022-08-16
1330 bug unable to refetch the data when fetching from graph fails

### LINKED ISSUE #1330: Bug: Unable to refetch the data when fetching from graph fails
## Steps To Reproduce

1. Turn off the internet connection, and we see the data that was fetching fails
2. on reconnecting to the internet, the re-fetching of data isn't happening.
3. only after refreshing the app does the data seem to be loading.

## The expected behavior
The application data needs to be fetched as soon as the mobile is connected to the internet.

Ekaanth:
fixed #1330 

---
  - https://app.asana.com/0/0/1202775146405616

### DIFF
--- a/src/hooks/useNetInfo.ts
+++ b/src/hooks/useNetInfo.ts
@@ -2,7 +2,7 @@ import NetInfo from '@react-native-community/netinfo';
 import {useEffect, useState} from 'react';
 
 export function useNetInfo() {
-  const [isConnected, setIsConnected] = useState(false);
+  const [isConnected, setIsConnected] = useState(true);
 
   useEffect(() => {
     const unsubscribe = NetInfo.addEventListener((state) => {
--- a/src/hooks/useNetInfo.ts
+++ b/src/hooks/useNetInfo.ts
@@ -2,7 +2,7 @@ import NetInfo from '@react-native-community/netinfo';
 import {useEffect, useState} from 'react';
 
 export function useNetInfo() {
-  const [isConnected, setIsConnected] = useState(true);
+  const [isConnected, setIsConnected] = useState(false);
 
   useEffect(() => {
     const unsubscribe = NetInfo.addEventListener((state) => {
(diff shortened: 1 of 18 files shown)


Ekaanth:
Follow up TODO: https://app.asana.com/0/1202739194556156/1202805065193803

src/hooks/useNetInfo.ts
```diff
@@ -2,7 +2,7 @@ import NetInfo from '@react-native-community/netinfo';
 import {useEffect, useState} from 'react';
 
 export function useNetInfo() {
-  const [isConnected, setIsConnected] = useState(false);
+  const [isConnected, setIsConnected] = useState(true);
```
ishrak-imam: Why would we want to set the initial state to`true`. In any case the initial state should be `false`. Is this fix related to the current scope of the bug?

src/ui/components/ScrollViewRefetch.tsx
```diff
@@ -12,8 +13,13 @@ type Props = Omit<ScrollViewProps, 'refreshControl'> & {
 
 export function ScrollViewRefetch({children, refetchQueries, ...props}: Props) {
   const {colors} = useTheme();
+  const {isConnected} = useNetInfo();
   const {refreshing, refetch} = useRefetch(refetchQueries);
 
+  useEffect(() => {
+    refetch();
+  }, [isConnected, refetch]);
+
```
ishrak-imam: This would be an incorrect place to implement this behaviour. The `ScrollViewRefetch` component is supposed to handle the pull to refresh only. New responsibilities are added to the component which are unrelated.

Another approach can look something like following.

```
function useApiRefetch() {
    const {isConnected} = useNetInfo();
    const {refetch} = useRefetch();

    useEffect(() => {
      if(isConnected){
        refetch();
      }
    }, [isConnected, refetch]);
}
```
And use this `useApiRefetch()` hook somewhere high up in the component chain. such as in the `AppNavigator` component.

ishrak-imam (approved):
Is the behaviour tested in both platforms? Let me know if I can help in iOS testing.

Ekaanth:
Please verify this on IOS, On android it works fine.

ishrak-imam (approved):
Works on iOS.

PR #1343, opened 2022-08-17
1316 refetch queries after sending a transaction

### LINKED ISSUE #1316: Refetch queries after sending a transaction
To make sure we are always displaying data that is up to date, let's make sure we are re-fetching queries after every transaction.

Ekaanth:
fixed #1316 

---
  - https://app.asana.com/0/0/1202777006571911

### DIFF
--- a/src/ui/components/AddBounty.tsx
+++ b/src/ui/components/AddBounty.tsx
@@ -21,7 +21,7 @@ type Props = {
 };
 
 export function AddBounty({onClose}: Props) {
-  const {data: bounty} = useBountiesSummary();
+  const {data: bounty, refetch: refetchBountiesSummary} = useBountiesSummary();
   const {formatBalance, stringToBn} = useFormatBalance();
 
   const [bountyTitle, setBountyTitle] = React.useState('');
@@ -72,6 +72,7 @@ export function AddBounty({onClose}: Props) {
       })
         .then(() => {
           snackbar('New bounty created');
+          refetchBountiesSummary();
         })
         .catch(() => {
           snackbar('Error while submitting bounty');
--- a/src/ui/components/AddBounty.tsx
+++ b/src/ui/components/AddBounty.tsx
@@ -15,16 +15,11 @@ import {BN, bnToHex} from '@polkadot/util';
 import {formattedStringToBn} from 'src/utils/balance';
 import {countUtf8Bytes} from 'src/utils';
 import {useStartTx} from 'context/TxContext';
-import {ApolloQueryResult} from '@apollo/client/core/types';
-import {SubstrateChainBounty} from 'src/generated/litentryGraphQLTypes';
+import {BountiesQueryResult} from 'src/api/hooks/useBounties';
 
 type Props = {
   onClose: () => void;
-  onSubmit: () => Promise<
-    ApolloQueryResult<{
-      substrateChainBounties: SubstrateChainBounty[];
-    }>
-  >;
+  onSubmit: () => BountiesQueryResult;
 };
 
 export function AddBounty({onClose, onSubmit}: Props) {
--- a/src/ui/components/AddBounty.tsx
+++ b/src/ui/components/AddBounty.tsx
@@ -15,13 +15,19 @@ import {BN, bnToHex} from '@polkadot/util';
 import {formattedStringToBn} from 'src/utils/balance';
 import {countUtf8Bytes} from 'src/utils';
 import {useStartTx} from 'context/TxContext';
+import {ApolloQueryResult} from '@apollo/client/core/types';
+import {SubstrateChainBounty} from 'src/generated/litentryGraphQLTypes';
 
 type Props = {
   onClose: () => void;
-  refetch: () => void;
+  onSubmit: () => Promise<
+    ApolloQueryResult<{
+      substrateChainBounties: SubstrateChainBounty[];
+    }>
+  >;
 };
 
-export function AddBounty({onClose, refetch}: Props) {
+export function AddBounty({onClose, onSubmit}: Props) {
   const {data: bounty} = useBountiesSummary();
   const {formatBalance, stringToBn} = useFormatBalance();
 
@@ -73,7 +79,7 @@ export function AddBounty({onClose, refetch}: Props) {
       })
         .then(() => {
           snackbar('New bounty created');
-          refetch();
+          onSubmit();
         })
         .catch(() => {
           snackbar('Error while submitting bounty');
--- a/src/ui/components/AddBounty.tsx
+++ b/src/ui/components/AddBounty.tsx
@@ -18,10 +18,11 @@ import {useStartTx} from 'context/TxContext';
 
 type Props = {
   onClose: () => void;
+  refetch: () => void;
 };
 
-export function AddBounty({onClose}: Props) {
-  const {data: bounty, refetch: refetchBountiesSummary} = useBountiesSummary();
+export function AddBounty({onClose, refetch}: Props) {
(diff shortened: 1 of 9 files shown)


src/ui/components/DelegateVoting.tsx
```diff
@@ -96,9 +96,11 @@ export function DelegateVoting({fromAccount, onClose}: DelegateVotingProps) {
           method: 'democracy.delegate',
           params: [delegatedAccount, conviction.value, bnToHex(delegateAmountBn)],
         },
+      }).then(() => {
+        refetchConvictions();
       });
     }
-  }, [delegatedAccount, conviction, delegateAmountBn, fromAccount?.address, startTx]);
+  }, [delegatedAccount, conviction, delegateAmountBn, fromAccount?.address, startTx, refetchConvictions]);
```
ishrak-imam: voting delegation doesn't affect conviction in any way. So no need to refetch conviction data after delegation.

Ekaanth: This component does't have conviction hook.

ishrak-imam: I can see the delegate voting component using the `useConvictions` hook but the point is the conviction data is not affected by the voting delegation operation. So no need to refetch the convictions.

Not sure what you mean @Ekaanth

src/ui/screens/Parachains/CrowdloanScreen.tsx
```diff
@@ -69,6 +69,7 @@ export function CrowdloanScreen({navigation}: ScreenProps) {
                 active={key === ongoingKey}
                 onPressContribute={() => {
                   setOpenContributeId(item.paraId);
+                  refetchCrowdloans();
```
silva-fj: When this item is pressed, no Tx is being sent.. This just opens the modal.

You need to fetch the query in the `ContributeBox` component. You can use [useLazyQuery](https://www.apollographql.com/docs/react/data/queries#manual-execution-with-uselazyquery) to fetch only after the Tx has resolved.

src/ui/components/AddBounty.tsx
```diff
@@ -72,6 +72,7 @@ export function AddBounty({onClose}: Props) {
       })
         .then(() => {
           snackbar('New bounty created');
+          refetchBountiesSummary();
```
silva-fj: What about the bounties list?

silva-fj:
It seems we are missing other cases in this PR, you can see the list of Tx we support here: `polkadot-api/src/txTypes.ts`

Ekaanth:
> txTypes

@franjs  Correct me if I'm wrong, but These are the types of transactions, but all the transaction types start from `startTx`

silva-fj:
> > txTypes
> 
> @franjs Correct me if I'm wrong, but These are the types of transactions, but all the transaction types start from `startTx`

That's correct. This is just for you to know more or less how many we have, you can search by the method

src/ui/screens/Parachains/CrowdloanScreen.tsx
Ekaanth: The `useCrowdloan` hook will get heavy if we start to add `useLazyQuery,` Thanks to @ishrak-imam  for helping me that I can pass refetch as a prop and invoke the query.

src/ui/screens/Council/CouncilScreen.tsx
```diff
@@ -330,7 +330,7 @@ function SubmitCandidacyModel({visible, setVisible, moduleElection}: SubmitCandi
             method: `${moduleElection.module}.submitCandidacy`,
             params: [argsLength === 1 ? [council?.candidates.length ?? 0] : []],
           },
-        });
+        }).then(() => refetchCouncil);
```
ishrak-imam: The function is not called.

```suggestion
        }).then(() => refetchCouncil());
```

src/ui/screens/Parachains/CrowdloanScreen.tsx
silva-fj: What do you mean by heavy?

Ekaanth: The purpose of the hook is just to fetch the active and ended, while adding lazyQuery will make it broader, Thereby we will have to create a new hook for lazyQuery, and since our case, we are rendering the same list, it would be additional work to invoke the lazyQuery client and assign the values back to the `useCrowdloan` variables since they are already declared as constants.

silva-fj: The idea is to expose another hook like `useLazyAllCrowdloans` or `useFetchAllCrowdloans` with `useLazyQuery` that makes use of the same query internally, you don't have to assign values or do anything else, the query will be cached as well

silva-fj: But in this case, passing the refetch as prop also works.

src/ui/screens/Parachains/CrowdloanScreen.tsx
```diff
@@ -190,10 +191,12 @@ function ContributeBox({
   visible,
   setVisible,
   parachainId,
+  refetch,
 }: {
   visible: boolean;
   setVisible: (_visible: boolean) => void;
   parachainId: string;
+  refetch: () => void;
```
silva-fj: This name is a bit confusing in the context of this component, you can call it `onContribute`. It does not need to know that will refetch some query

src/ui/components/AddBounty.tsx
```diff
@@ -18,10 +18,11 @@ import {useStartTx} from 'context/TxContext';
 
 type Props = {
   onClose: () => void;
+  refetch: () => void;
```
silva-fj: This is the same case, you can use `onSubmit` or `on(Add/Propose)Bounty`

src/ui/screens/Bounty/BountiesScreen.tsx
```diff
@@ -20,12 +20,16 @@ const ItemSeparator = () => <Padder scale={0.5} />;
...
 export function BountiesScreen({navigation}: ScreenProps) {
   const {openBottomSheet, closeBottomSheet, BottomSheet} = useBottomSheet();
-  const {data: bounties, loading} = useBounties();
+  const {data: bounties, loading, refetch: refetchBounties} = useBounties();
 
   const toBountyDetails = (index: string) => {
     navigation.navigate(bountyDetailScreen, {index});
   };
 
+  const bountiesRefetch = () => {
```
silva-fj: Why this extra function?

silva-fj: If the issue was a typescript error, then lets put the function back to avoid coupling the type with Apollo, that's too specific for a generic prop

Ekaanth: I have updated the code now, it decoupled and the type is added to the hook.

src/ui/components/AddBounty.tsx
```diff
@@ -15,12 +15,14 @@ import {BN, bnToHex} from '@polkadot/util';
 import {formattedStringToBn} from 'src/utils/balance';
 import {countUtf8Bytes} from 'src/utils';
 import {useStartTx} from 'context/TxContext';
+import {BountiesQueryResult} from 'src/api/hooks/useBounties';
 
 type Props = {
   onClose: () => void;
+  onSubmit: () => BountiesQueryResult;
```
silva-fj: This is the same as having `refetch` as prop name, we cannot pass anything that is not a function that returns a query result if in the future we want to add something. Just keep the `() => void` as you had before

silva-fj (approved):

PR #1356, opened 2022-08-29
issue-1342: closeExternalAccount manually

### LINKED ISSUE #1342: Bug: Add external account bottom sheet open.
## Steps To Reproduce

1. Press on add external account
2. Add an external account via QR or type-in
3. and press confirm => press confirm and press close.
4. Press on the added account and press back button
5. add external account bottom sheet is still open.

## The expected behavior

The bottom sheet has to be closed when we return to my accounts page.

Ekaanth:
fixed #1342 

---
  - https://app.asana.com/0/0/1202837001479663

### DIFF
--- a/src/ui/screens/AccountsScreen.tsx
+++ b/src/ui/screens/AccountsScreen.tsx
@@ -56,6 +56,7 @@ export function AccountsScreen({navigation}: Props) {
     navigation.setOptions({
       headerRight: HeaderRight,
     });
+    // TODO: https://github.com/gorhom/react-native-bottom-sheet/issues/972
     closeExternalAccount();
   }, [navigation, HeaderRight, closeExternalAccount]);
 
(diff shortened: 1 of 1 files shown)


src/ui/screens/AccountsScreen.tsx
```diff
@@ -56,7 +56,8 @@ export function AccountsScreen({navigation}: Props) {
     navigation.setOptions({
       headerRight: HeaderRight,
     });
-  }, [navigation, HeaderRight]);
+    closeExternalAccount();
```
silva-fj, line 60: Please, double check that there are not other issues introduced with this. Also, can you create an issue and add a TODO here with the link to keep track of it (If the library fixes the bug this should be removed)

Ekaanth: Ok, I'll verify that and add a TODO in the next commit.

silva-fj (approved):

